PR #11, opened 2018-10-30
added comments on installation and running

BenjaminRodenberg:
Some comments on installation and running of the systemtests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ see ```python3 local_test.py --help``` for explanation of the command line argum
 ## Troubleshooting
 
 * If you receive an error like ```W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/xenial/InRelease  Temporary failure resolving 'archive.ubuntu.com'```, the following [answer on stackoverflow](https://stackoverflow.com/a/40516974) might help. Even, if the suggested verification step does not work, give the systemtests another try. Sometimes it works...
-* If you run into problems during compilation of preCICE in your docker container, use ```force_rebuild=True```.
+* If you run into problems during compilation of preCICE in your docker container, try to use ```python3 local_test.py --force_rebuild precice```.
 
 # preCICE
 - Dockerfile.precice docker image with ubuntu 16.04 and preCICE
--- a/README.md
+++ b/README.md
@@ -15,30 +15,13 @@ Make sure to install
     * with ```configparser``` (install for example with ```pip3 install configparser```)
 * additional dependencies depending on the test you want to run
 
-## Running 
-
-### Run a system test
+## Run a system test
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,50 @@
...
+```python3 local_test.py -b mpi_single_ports -s of-of su2-ccx -f precice```
+
+see ```python3 local_test.py --help``` for explanation of the command line arguments.
+
+### Build & Run preCICE image interactively
+
+**Build:**
+
+```
+python3 -c 'import docker; docker.build_image("precice-interactive", "Dockerfile.precice", force_rebuild = False)'
```
floli: This command uses internal stuff and may break without any notice. Anyway, why do you use it at all? `local_test.py` builds the precice image just fine.

If you want to build your own precice image use the standard docker commands `docker build -t precice/precice:develop -f Dockerfile.precice --build-arg ...`

README.md
```diff
@@ -1,6 +1,50 @@
...
+```
+
+*comment:* if already a previous build of the preCICE image exists you can use ```force_rebuild = True``` to make sure that the most recent precice version is used.
+
+**Run:**
+
+* Make sure that your image has been build via ```docker images```.
+* Run the image in interactive mode via ```docker run -ti precice-interactive /bin/bash```.
```
floli: Maybe we should provide an official docker image, but the Dockerfile here in the systemtest repo is not intended for general use (albeit it *may* work just fine)

floli (changes requested):
Please make sure, it's clear that systemtests are for internal tests and nothing like the official docker image.

Still, thanks for the documentation additions!

README.md
BenjaminRodenberg: Should we entirely remove the paragraph "Build & Run preCICE image interactively" in the README.md and provide an official preCICE docker image instead?

README.md
```diff
@@ -1,6 +1,33 @@
...
+Example command:
+
+```python3 local_test.py -b mpi_single_ports -s of-of su2-ccx -f precice```
+
+see ```python3 local_test.py --help``` for explanation of the command line arguments.
+
+## Troubleshooting
+
+* If you receive an error like ```W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/xenial/InRelease  Temporary failure resolving 'archive.ubuntu.com'```, the following [answer on stackoverflow](https://stackoverflow.com/a/40516974) might help. Even, if the suggested verification step does not work, give the systemtests another try. Sometimes it works...
+* If you run into problems during compilation of preCICE in your docker container, use ```force_rebuild=True```.
```
floli: Rather `--force_rebuild precice` as argument to `local_test.py`

PR #21, opened 2019-01-05
Bring back travis integration

shkodm:
#### General approach: 

* All tests are run on Ubuntu 16.04 ( Ubuntu 18.04 will come in separate PR)
* During "Build stage", preCICE images are built and pushed to the [dockerhub](https://hub.docker.com/u/precicecoupling). 
* During "System tests" stage, all tests are executed, and results are pushed  to the [ precice_st_output ](https://github.com/precice/precice_st_output) repository. In case of succesful execution,  only the log file, with name of the system_test, time, commit hashes, etc is pushed. The clickable link to the travis job  is in the commit message, for the reference. 
  If results are different from the `referenceOutput` also folder with results is commited.  If the job fails during building of the adapter we detect it and write corresponding commit message as well. 

  In order to distinguish between different cases that fail, result are pushed either to 'Ubuntu1604' 
  or 'Ubuntu1804' folder. Inside those, the  folder with the output has name as 'Output_$system_test_name'  
  If the next build fixes previously broken test, output folder is deleted. Similarly, it is ensured, that different    wrong results are not mixed. ( E.g. if the tests fails, produces wrong output and the next commit produces *different* wrong output, only latest wrong output stays in the repository ). 

* The tests should probably be run nightly as a cron job

 You have a look at  repository with the output to get a feeling of the structure [in this repo]( https://github.com/precice/precice_st_output ). 
Sample job build ( I already enabled  Travis to test on other branch ) is [here](https://travis-ci.org/precice/systemtests/builds/475788595).

### DIFF
--- a/common.py
+++ b/common.py
@@ -9,7 +9,10 @@ def call(cmd, **kwargs):
 def ccall(cmd,  **kwargs):
     """ Runs cmd in a shell, returns its return code. Raises exception on error """
     return call(cmd, check = True, **kwargs)
-    
+
+def capture_output(cmd, **kwargs):
+    """ Runs cmd in a shell and captures output """
+    return subprocess.run(cmd, stdout=subprocess.PIPE, **kwargs).stdout.decode('utf-8')
 
 import os
 
--- a/push.py
+++ b/push.py
@@ -1,7 +1,7 @@
 """Script for pushing output and logfile to a repository.
 
 This script pushes output files of a system test, if they exists, and a logfile
-to the repository: https://github.com/kunstrasenspringer/precice_st_output.
+to the repository: https://github.com/precice/precice_st_output.
 
     Example:
         Example use to push output files and logfile of system test of-of:
@@ -10,13 +10,16 @@ to the repository: https://github.com/kunstrasenspringer/precice_st_output.
 """
 
 import argparse, os, sys, time
-from common import ccall
+from common import ccall, capture_output
 
 # Parsing flags
 parser = argparse.ArgumentParser(description='Build local.')
 parser.add_argument('-b', '--branch', help="log choosen preCICE branch")
 parser.add_argument('-t', '--test', help="choose system tests you want to use")
 parser.add_argument('-s', '--success', action='store_true' ,help="only upload log file")
+parser.add_argument('-o', '--os', type=str,help="Version of Ubuntu to use", choices =
+            ["1804", "1604"], default= "1604")
+
 args = parser.parse_args()
 
 if __name__ == "__main__":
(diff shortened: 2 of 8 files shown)


common.py
```diff
@@ -9,7 +9,10 @@ def call(cmd, **kwargs):
 def ccall(cmd,  **kwargs):
     """ Runs cmd in a shell, returns its return code. Raises exception on error """
     return call(cmd, check = True, **kwargs)
-    
+
+def capture_output(cmd, **kwargs):
+    """ Runs cmd in a shell and captures output """
+    return subprocess.run(cmd, stdout=subprocess.PIPE, **kwargs).stdout.decode('utf-8')
```
floli, line 15: Better use decode without arguments, which uses the system's default local.

push.py
```diff
@@ -10,29 +10,33 @@
...
 import argparse, os, sys, time
-from common import ccall
+from common import ccall, capture_output
 
 # Parsing flags
 parser = argparse.ArgumentParser(description='Build local.')
 parser.add_argument('-b', '--branch', help="log choosen preCICE branch")
 parser.add_argument('-t', '--test', help="choose system tests you want to use")
 parser.add_argument('-s', '--success', action='store_true' ,help="only upload log file")
+parser.add_argument('-o', '--os', type=str,help="Version of Ubuntu to use", choices =
```
floli, line 20: Please implement that in a generic way, not only for Ubuntu, there are other Linux versions in existence.

push.py
```diff
@@ -47,37 +51,56 @@
...
     localtime = str(time.asctime(time.localtime(time.time())))
     log.write("System testing at " + localtime + "\n")
     log.close()
 
     # Pushing outputfiles and logfile to repo.
     # Clone repository.
-    ccall("git clone https://github.com/kunstrasenspringer/precice_st_output")
-    os.chdir(os.getcwd() + "/precice_st_output")
+    ccall("git clone https://github.com/precice/precice_st_output")
+    log_dir = os.getcwd() + "/precice_st_output/Ubuntu" + args.os
```
floli, line 63: Same here, no hardcoded Ubuntu please.

system_testing.py
```diff
@@ -25,7 +26,15 @@ def build(systest, tag, branch, local, force_rebuild):
                            build_args = {"from" : docker.get_namespace() + baseimage_name},
                            force_rebuild = force_rebuild)
     else:
-        docker.build_image(tag = test_tag, force_rebuild = force_rebuild)
+        match = re.search(r'ubuntu(\d+)', baseimage_name)
```
floli, line 29: No hardcoded Ubuntu, just reference to the filename of the base docker file.

PR #22, opened 2019-01-06
Allow adapter to trigger custom systemtests build 

shkodm:
**NOTICE**: This should be merged (hopefully) only after #21.  Also one should probably read that PR, before reading this. 

## Problem

 Even when having daily cron-job based builds with Travis several problems might occur. What if in one day several commits were pushed to adapters ( or maybe also preCICE itself ). I know, I know, some adapters are not in active development, but from what I can see some are (e.g.  [openfoam-adapter](https://github.com/precice/openfoam-adapter) ). Then daily build fails, wrong results or failure during building is produced and one has to manually dig through the series of changes in the adapter to find out which one exactly caused the problem. This is not nice. 

Reasoning above should suggest, that some kind of more continuous integration should exist for adapters also ( at least for the master branch ). But we already have the tests for adapters in this repository, that are integrated nicely with Travis ( after #21 ). Can we somehow reuse this functionality for adapters? Of course, we don't want to run full systemtests suite, with all the adapters and additional building of preCICE, but lets say we just use already pushed to dockerhub preCICE image (so that tests will be executed fast enough) and only tests that are related to the adapter are run (e.g. for OpenFOAM we test OpenFOAM-OpenFOAM and OpenFOAM-Calculix, but for SU2 just SU2-Calculix ). How to achieve it with least effort, no code duplication and make sure it is robust and useful? 

## Solution 

Turns out we can use [Travis API v3]( https://docs.travis-ci.com/user/triggering-builds/) to make dependent builds, i.e. we can trigger `systemtests` Travis build during execution of Travis build of e.g. openFOAM adapter. This way we can make integration *continuous*, so new commits on adapters are checked. Now, to make sure, with this customly triggered builds, that only set of tests related to adapter is executed and we do not do any extra work, we can use more features of Travis API, namely use custom job bodies ( and send them via API request as JSON). However, this is not enough, since the maintainers of the adapter ( and the 'passing' or 'fail' badge in the `README.md`...) need to be informed, and we want to keep history of failures of the adapters in its own Travis page. To do this we can again use dependent builds, now making systemtests build trigger custom build of the adapter, that just fails, if the systemtests failed. This way adapter gets to know that it failed, and if desired can find reference output in the [precice_st_output](https://github.com/precice/precice_st_output)

Since build configuration of this custom Travis jobs are pretty similar, with only difference being names of tests to run, names of repositories, etc., we can write general python script that can send triggering of different systemtests (and corresponding failure callback) from different adapter and allows easy addition of new ones. 

Then the only thing remaining is to add `.travis.yml`  to the adapter, which (excluding system specs can be very simple) 
```yml
    - if: branch = master
      script: |
        curl -LO --retry 3 https://raw.githubusercontent.com/precice/systemtests/master/trigger_systemtests.py
        python trigger_systemtests.py --adapter openfoam
``` 
And of course, we need to add `TRAVIS_ACCESS_TOKEN` in each adapter repository as a secure variable, to be used as authentication when triggering Travis with API. But this is not a big price to pay. 

To illustrate, the solution above, following will be executed on commit to OpenFOAM adapter master branch: 

1) Travis starts building openFOAM adapter, send trigger request to systemtests and exits with success 
image
                                                               *openfoam-adapter travis page*

2) Travis starts building systemtests using custom job recipe, that was just sent, runs OpenFOAM-OpenFOAM, OpenFOAM-CalculiX, if everything is fine exits with success, otherwise pushes wrong output and logfile to the output repository and triggers custom Travis build of OpenFOAM
image
                                                 *systemtests travis page (note that only openfoam related tests  are run)* 

3) (If systemtests failed) Travis starts building OpenFOAM, and since the request was just to fail, it fails. Fin. 
image
                                                             *openfoam-adapter travis page*

In this PR I add `trigger_systemtests.py`, that does all of the above. I already tested it on my fork of the Calculix, OpenFOAM, and SU2 adapters. Discussion and suggestions what else can be improved are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,3 +65,9 @@ Python script to compare reference data with output data.
 2. Create a `Dockerfile` in there.
 3. Create a directory `referenceOutput` there.
 4. Copy the output files to a folder `/Output/` (inside the container).
+
+# Adding CI to a new adapter
+
+As described in [#22](https://github.com/precice/systemtests/pull/22),  using `trigger_systemtests.py` it is possible to trigger custom systemtests from the travis job of the another adapter, therefore providing continous integration for the adapter on each commit to a particular branch. If you want to add another adapter, or modify tests that are run for it, you need to modify `nm_repo_map` and `nm_test_map` variables in the `trigger_systemtests.py`, that describe mapping between adapter name and repository and adapter name and set of test cases correspondingly.
+
+The adapter's Travis build would be considered successful if systemtests Travis build exits with status 'Failed' or 'Errored'. It would exit with 'Success', if the systemtests Travis build exits with 'Success' or 'Canceled'.
--- a/trigger_systemtests.py
+++ b/trigger_systemtests.py
@@ -7,6 +7,8 @@
 
 import json
 import os
+import time
+from sys import exit
 import argparse
 from string import Template
 from urllib.parse import urlencode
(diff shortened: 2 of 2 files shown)


MakisH (changes requested):
Two notes:
* If I understand correctly, this would lead to an "always success", potentially followed by a failure after a while. So, we should keep in mind that the "success marks" on the adapters may be temporarily false. Could we somehow improve this?
* Add a note on README for the additional steps required by this PR when adding a new system test (i.e. how to map tests to adapters).

shkodm:
@MakisH , 
> If I understand correctly, this would lead to an "always success", potentially followed by a failure after a while. So, we should keep in mind that the "success marks" on the adapters may be temporarily false. Could we somehow improve this? 

So now that I am thinking about it, there can be another approach. We can 
* Start a travis job from the adapter, trigger systemtests, but do not exit immediately, but enter some kind of a while loop like 
```python
while systemTestStatus == "running":
      systemTestStatus = querySystemtests()
return systemTestStatus 
```
This way the job on the adapter will exit only after the job on systemtests finished. This way we will probably have cleaner Travis history, but also first job would be just wasting computing time. I can implement it anyway, just to see how it would work and explore the options.

MakisH:
Can we maybe exit the first run with something else rather than "success" or "failure"? E.g. "interrupted"?

shkodm:
As discussed with @MakisH in person, we decided to add both options. Then each adapter can choose what is most suitable.  Thus I implemented, what I described two comments above and it can be switched on with `--wait` argument. Thus to use it in the adapter:
```yaml
   if: branch = master
      script: 
        - curl -LO --retry 3 https://raw.githubusercontent.com/precice/systemtests/master/trigger_systemtests.py
        - travis_wait 60 python trigger_systemtests.py --adapter openfoam --wait
```
`travis_wait` is used so Travis will not kill our job because of no output. I tested on my adapter forks and everything seems to be working as intended.

MakisH:
Please also have https://github.com/precice/systemtests/issues/25 in mind. But feel free to deal with the Ubuntu-specific issue alltogether in a separate Pull Request.

Issue #14, opened 2018-11-13
RFC: Test multiple variants


floli:
Mutliple preCICE Dockerfile can be specified:
```
Dockerfile.Ubuntu1804
Dockerfile.Ubuntu1804.Boost160
Dockerfile.Ubuntu1804.NoPETSc
Dockerfile.Ubuntu1604
Dockerfile.Ubuntu1604.Boost160
```
these are matched with `Test_` directories:

```
Test_of-of
Test_of-of.Ubuntu1804
Test_of-of.Ubuntu1804.Boost160
Test_su2.Ubuntu1604
```
Matching logic:

* For each preCICE Dockerfile, there should be maximum one match in each test class (class = {of-of, su2} ).
* `Test_className` is a default
* If a `Dockerfile.A.B.C` is not matched within a class, it is right stripped of one component and matched again. If only one component (`A`) is left, the default `Test_className` is picked. If no default exists, no test is executed.

Examples:
* `Dockerfile.Ubuntu1804` -> { `Test_of-of.Ubuntu1804`, (no matching su2) }
* `Dockerfile.Ubuntu.1804.Boost160` -> {`Test_of-of.Ubuntu1804.Boost160`}
* `Dockerfile.Ubuntu.1804.NoPETSc` -> {`Test_of-of.Ubuntu1804`}
* `Dockerfile.Ubuntu1604` -> { `Test_of-of`, `Test_su2.Ubuntu1604` }
* `Dockerfile.Ubuntu1604.Boost160` -> {`Test_of-of`, `Test_su2.Ubuntu1604`}

BenjaminRodenberg:
If I run on 1c98b2afa94c7a93a13eed7c17b71d68f74301a2 `python3 local_test.py --systemtest of-of --dockerfile Dockerfile.Ubuntu1804 --force-rebuild precice`, I get an error. Here, appropriate specialization of the Dockerfiles for different Ubuntu versions is probably needed.

With 16.04 everything works `python3 local_test.py --systemtest of-of --dockerfile Dockerfile.Ubuntu1604
`

BenjaminRodenberg:
https://github.com/precice/systemtests/commit/6c9e1cfae8e0bfb83518f51d4f36a0a56b619f4a allows to run the following commands

```
python3 local_test.py -s bindings -d Dockerfile.Ubuntu1804
python3 local_test.py -d Dockerfile.Ubuntu1804
python3 local_test.py -d Dockerfile.Ubuntu1604
```

[...]

PR #27, opened 2019-02-07
Remove hardcoding of Ubuntu

### LINKED ISSUE #25: Do not hardcode Ubuntu
The recent pull request #21 hard coded ubuntu in the code.

Hard coding for a single Linux distribution is not the way to go and we eventually want to test on/for other distributions as well. Please remove it.

shkodm:
Resolves #25 
On Travis we still test only on it (16.04 and 18.04) , so I keep it as only choices for `system_testing.py` arguments.

### DIFF
--- a/push.py
+++ b/push.py
@@ -17,7 +17,7 @@ parser = argparse.ArgumentParser(description='Build local.')
 parser.add_argument('-b', '--branch', help="log choosen preCICE branch")
 parser.add_argument('-t', '--test', help="choose system tests you want to use")
 parser.add_argument('-s', '--success', action='store_true' ,help="only upload log file")
-parser.add_argument('-o', '--os', type=str,help="Variant used", default= "Ubuntu1604")
+parser.add_argument('--base', type=str,help="Base preCICE image used", default= "Ubuntu1604")
 
 args = parser.parse_args()
 
@@ -26,7 +26,7 @@ if __name__ == "__main__":
 
     # Creating new logfile. if it exists, truncate content.
     log = open("log_" + systest, "w")
-    foam_version = 4.1 if args.os == "1604" else 5
+    foam_version = 4.1
     # Saving versions of used software in system test systest.
     if systest == "of-ccx":
         log.write("OpenFOAM version: {}\n".format(foam_version))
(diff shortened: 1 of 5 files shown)


push.py
```diff
@@ -17,8 +17,8 @@
 parser.add_argument('-b', '--branch', help="log choosen preCICE branch")
 parser.add_argument('-t', '--test', help="choose system tests you want to use")
 parser.add_argument('-s', '--success', action='store_true' ,help="only upload log file")
-parser.add_argument('-o', '--os', type=str,help="Version of Ubuntu to use", choices =
-            ["1804", "1604"], default= "1604")
+parser.add_argument('-o', '--os', type=str,help="Distribution to use", choices =
```
floli: Is that really a distribution or the named variant of the base docker file? Do you adhere to https://github.com/precice/systemtests/issues/14?

floli (changes requested):

push.py
shkodm: I read #14, but here the same argument as for `system_testing.py` applies - right now we don't test anything else, but  'Ubuntu1604' and 'Ubuntu1804' on Travis, and I believe with @MakisH  we concluded that it is enough to test on different distributions. So `distributions` makes more sense to me.
 But if we decide to test and push results from different variants  ( as in #14 ) based on the same operating system, I would change this bit.

floli: Well, if we don't have the infrastructure in place we will never be able to test something else, therefore we don't do it.

We should definitely be able to test on arbitrary base images, as laid out in #14. If you feel that this scheme is not suitable, please comment here.

For travis only testing Ubuntu is ok, but the test suite should test other variants as well. None of supercomputers use Ubuntu.

MakisH: I confirm @shkodm that we agreed (between the two of us) that we need to be able on different distributions and that for Travis Ubuntu is enough. I am not sure I understand the problem here and what should be done differently.

shkodm: @floli, Allright, you are right, I will adjust that bit.

floli:
What is the status of this PR? I have added an Arch based preCICE image and will eventually also try to add solver testing images for it. As far as I see, `push.py` is still Ubuntu only.

floli (approved):

PR #29, opened 2019-02-14
Update  Calculix to 2.15

### LINKED ISSUE #28: Update all CCX systemtests
We just updated the CalculiX adapter to v2.15, see https://github.com/precice/calculix-adapter/commit/e212e16b4392958f632b12265c772b08c792f35c. 

As expected, the system tests now fail. The used CalculiX version here also needs to be updated. Maybe also reference files need to be updated.

shkodm:
Fixes #28. Results seems to be unchanged.


MakisH:
I guess this is failing by definition, but everything should be ok. Merging.

PR #30, opened 2019-02-16
Fix C/CPP Include ENV not set

fsimonis:
This _should_ fix the currently failing build of the python bindings.


fsimonis:
I'll resolve this in another PR

PR #31, opened 2019-02-20
WIP Fix the Python bindings for preCICE v1.4.0

MakisH:
See https://github.com/precice/precice/issues/220#issuecomment-465312880


fsimonis:
This is not required once we set the ENV correctly.
See:
*  #39 for ENV
*  #40 for how to pip can install it without any additional arguments

PR #39, opened 2019-02-27
Shrink the image size and reduce layers

fsimonis:
This PR reduces the docker image size and layer counts. It will:
* add clean-up after apt-get
* removes the boost build directory after install in Ubuntu 16.04
* removes `.o`and `.os` files after building precice with `scons`
* merges `ENV` calls
* merges pip calls
* adds  `CPATH` and `CPLUS_INCLUDE_PATH` to include precice turning `-I` in solverdummies obsolete
* makes the precice path explicit in clone.

reviewers: last review please merge if ok.


fsimonis:

PR #40, opened 2019-02-27
WIP Add PIP install for the bindings

fsimonis:
This PR:
* installs only `python[2|3]-pip` via `apt-get`.
* installs all the dependencies via `pip` into the user (which will work for actual users)
* installs the precice bindings directly via `pip`

The second point should become obsolete, once we add correct dependencies to the precice package.

This pr depends on #39 for the env changes
reviewers: please comment /  approve only


fsimonis:

PR #41, opened 2019-02-27
Add preCICE include path to ENV

fsimonis:
This is essentially a #39 compatible version of #30.


fsimonis:

PR #43, opened 2019-02-27
Shrink the image size and reduce layers

fsimonis:
Proxy for #39

### DIFF
--- a/Dockerfile.Ubuntu1604
+++ b/Dockerfile.Ubuntu1604
@@ -30,10 +30,10 @@ RUN wget -nv 'https://dl.bintray.com/boostorg/release/1.65.1/source/boost_1_65_1
 ARG CACHEBUST
 
 # Setting some environment variables for installing preCICE
-ENV CPLUS_INCLUDE_PATH="$CPLUS_INCLUDE_PATH:/usr/include/eigen3"
-ENV CPATH="/usr/include/eigen3:${CPATH}"
-ENV PETSC_DIR="/usr/lib/petscdir/3.6.2/"
-ENV PETSC_ARCH="x86_64-linux-gnu-real"
+ENV CPLUS_INCLUDE_PATH="$CPLUS_INCLUDE_PATH:/usr/include/eigen3" \
+    CPATH="/usr/include/eigen3:${CPATH}" \
+    PETSC_DIR="/usr/lib/petscdir/3.6.2/" \
+    PETSC_ARCH="x86_64-linux-gnu-real"
 
 # Building preCICE
 ARG branch=develop
@@ -51,7 +51,7 @@ RUN scons petsc=$petsc_para mpi=$mpi_para python=$python_para -j$(nproc) && \
 
 # Setting preCICE environment variables
 ENV PRECICE_ROOT="/precice"
-ENV LD_LIBRARY_PATH="$PRECICE_ROOT/build/last:${LD_LIBRARY_PATH}"
-ENV LIBRARY_PATH="$PRECICE_ROOT/build/last:${LIBRARY_PATH}"
-ENV CPATH="$PRECICE_ROOT/src:${CPATH}"
-ENV CPLUS_INCLUDE_PATH="$PRECICE_ROOT/src:${CPLUS_INCLUDE_PATH}"
+ENV LD_LIBRARY_PATH="$PRECICE_ROOT/build/last:${LD_LIBRARY_PATH}" \
+    LIBRARY_PATH="$PRECICE_ROOT/build/last:${LIBRARY_PATH}" \
+    CPATH="$PRECICE_ROOT/src:${CPATH}" \
+    CPLUS_INCLUDE_PATH="$PRECICE_ROOT/src:${CPLUS_INCLUDE_PATH}"
--- a/Dockerfile.Ubuntu1604
+++ b/Dockerfile.Ubuntu1604
@@ -46,7 +46,8 @@ WORKDIR precice
 ARG petsc_para=no
 ARG mpi_para=yes
 ARG python_para=no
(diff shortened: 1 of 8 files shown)


Dockerfile.Ubuntu1604
```diff
@@ -14,38 +14,44 @@ RUN apt-get -qq update && apt-get -qq install \
...
 # Some parameters for the build, you can set them in the build command e.g.
 # sudo docker build Dockerfile.precice --build-arg petsc_para=yes --build-arg mpi_para=yes .
 # this will result in
 # scons petsc=yes mpi=yes python=no compiler="mpicxx" -j2
 ARG petsc_para=no
 ARG mpi_para=yes
 ARG python_para=no
-RUN scons petsc=$petsc_para mpi=$mpi_para python=$python_para -j$(nproc)
+RUN scons petsc=$petsc_para mpi=$mpi_para python=$python_para -j$(nproc) && \
+    find . -type f \( -name "*.o" -or -name "*.os" \) -exec rm {} +
```
floli, line 51: Please add a comment, why this is necessary. Don't think it's obvious that's just for reducing image size.

Dockerfile.Ubuntu1604
```diff
@@ -14,38 +14,44 @@ RUN apt-get -qq update && apt-get -qq install \
     python-numpy \
     python-dev \
     wget \
-    bzip2
+    bzip2 && \
+    rm -rf /var/lib/apt/lists/*
```
MakisH, line 18: Why not do a normal `apt-get clean`?

Dockerfile.Ubuntu1804
```diff
@@ -15,31 +15,35 @@ RUN apt-get -qq update && apt-get -qq install \
     python-numpy \
     python-dev \
     wget \
-    bzip2    
+    bzip2 && \
+    rm -rf /var/lib/apt/lists/*
```
MakisH, line 19: Same here about `apt-get clean`.

Test_bindings.Ubuntu1804/Dockerfile
```diff
@@ -9,7 +9,8 @@ ENV USER=root
 
 # Installing necessary dependecies
 RUN apt-get -qq update && apt-get -qq install \
-    cython python-numpy python-enum34 python-mpi4py cython3 python3-numpy python3-mpi4py
+    cython python-numpy python-enum34 python-mpi4py cython3 python3-numpy python3-mpi4py && \
+    rm -rf /var/lib/apt/lists/*
```
MakisH, line 13: Same here about `apt-get clean`.

Test_bindings/Dockerfile
```diff
@@ -9,13 +9,14 @@ ENV USER=root
 
 # Installing necessary dependecies
 RUN apt-get -qq update && apt-get -qq install \
-    python-pip python-enum34 python3-pip
+    python-pip python-enum34 python3-pip && \
+    rm -rf /var/lib/apt/lists/*
```
MakisH, line 12: Well, same everywhere (`apt-get clean`).

MakisH (changes requested):
My only requested change is using `apt-get clean` instead of manually deleting files.

Apart from this, we could use [Apline Linux](https://coderwall.com/p/s4ofoq/docker-why-i-prefer-alpine-as-base-instead-of-ubuntu) instead of Ubuntu to really reduce the image size. But then we would not reproduce an Ubuntu testing environment, of course. However, it could be useful if one day we wanted to ship Docker images.

Dockerfile.Ubuntu1604
fsimonis: True, alternatively we could wipe the entire directory apart from the library and executables.

Dockerfile.Ubuntu1604
fsimonis: Ubuntu Docker images run `apt-get clean` after each install. [More here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers)

Dockerfile.Ubuntu1604
fsimonis: I added a comment regarding the clean-up.
Feel free to resolve the conversation if you are ok with it.

fsimonis:
@MakisH I don't think Alpine is a good choice here. We try to model a plausible environment, so the Ubuntu images may be the best choice.

MakisH:
Interesting piece of information, for any other projects trying to reduce their image size as well:

> In Docker 1.10 and higher, only the instructions RUN, COPY, ADD create layers. Other instructions create temporary intermediate images, and do not directly increase the size of the build.

So, at the end, merging the `ENV` may not have been necessary.

From the (linked above by @fsimonis ) [Docker best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers).

**Edit:** actually, this is reverted further below:

> Each ENV line creates a new intermediate layer, just like RUN commands. This means that even if you unset the environment variable in a future layer, it still persists in this layer and its value can be dumped.

MakisH:
Travis error:

```
  File "/usr/bin/pip3", line 9, in <module>
    from pip import main
ImportError: cannot import name 'main'
The command '/bin/sh -c pip2 install --upgrade pip &&     pip2 install Cython mpi4py &&     pip3 install --upgrade pip &&     pip3 install Cython mpi4py numpy' returned a non-zero code: 1
```

fsimonis:
@MakisH I merged the ENV to prevent the generation of the intermediate layers. My hope is that docker has less to do this way.

shkodm (approved):

MakisH (approved):

Issue #51, opened 2019-03-05
Create systemtest baseimage using the preCICE Release Binaries


BenjaminRodenberg:
From preCICE v1.4.0 we are planning to provide binaries. We should not forget to test them in our systemtests. A new baseimage like [`Dockerfile.Ubuntu1604`](https://github.com/precice/systemtests/blob/master/Dockerfile.Ubuntu1604) or [`Dockerfile.Ubuntu1804`](https://github.com/precice/systemtests/blob/master/Dockerfile.Ubuntu1804) will be required.

BenjaminRodenberg:
First suggestion for a naming convention (compare #14):

* rename current `Dockerfile.Ubuntu1604` to `Dockerfile.Ubuntu1604.CMake`
* create new `Dockerfile.Ubuntu1604.Debian_v1_4_0`

These names are for sure not perfect and open to discussion.

BenjaminRodenberg:
What's the status on this issue? It would be good to get this done before the next release, I assume.

PR #54, opened 2019-03-07
WIP: Port all baseimages to cmake and use the different installation procedures

BenjaminRodenberg:
We should test preCICE being installed via the different available installation procedures. These are:

* use `make install` and install in user directory (`INSTALL = home`)
* use `sudo make install` and install in system directory (`INSTALL = sudo`)
* use `make package` to create debian package and install it (`INSTALL = package`) *comment:* this is not identical to #51. In #51 we want to use the binary provided in the release; here, we build the binary on our own.

For the different cases we have to provide different docker base images `Dockerfile.OS.INSTALL`. Example: `Dockerfile.Ubuntu1804.sudo` is based on Ubuntu18.04 and installs preCICE using cmake in a system directory using sudo.

Currently there are a few open issues:

* [x] **(1) Treat user rights properly.** We create a user `precice`, since using the `root` user is not recommended (pip, mpiexec). However, for some use cases `precice` needs sudo rights (install debian package). Currently, we just add `precice` to the `sudo` group for these cases. This does not look correct, but we also want to avoid creating big layers, by switching between users for the installation. Is there a proper way to solve this problem?
**Example:**
In `Dockerfile.Ubuntu1604.home` the user `precice` does not need `sudo` 
https://github.com/precice/systemtests/blob/dcb07b6a59c39dc6d56750faf6aefb2ba50c69d4/Dockerfile.Ubuntu1604.home#L36-L39
In `Dockerfile.Ubuntu1604.home` the user `precice` needs `sudo` 
https://github.com/precice/systemtests/blob/dcb07b6a59c39dc6d56750faf6aefb2ba50c69d4/Dockerfile.Ubuntu1604.sudo#L36-L39
* [ ] **(2) User `precice` should be used in the tests, as well** Currently, this is not consistent.
* [ ] **(3) Reduce layer size** This is especially important with respect to the aforementioned todo **(1)**


BenjaminRodenberg:
We should also make sure that the actual tests (e.g. `Test_bindings`) are run on all available base images. This checks that the different installation procedures provide a properly working preCICE. However, I would not address this issue in this PR. This is related to #14.

BenjaminRodenberg:
With the current state of `Dockerfile.OS.package` I am fine and I would leave it as it is. For `Dockerfile.OS.sudo`, we should still try to reduce the layer size.

fsimonis:
I ran the tests locally using the sudo base image variants and fixed a few things:
* The `FORTRAN` env now gets set to `gfortran` in the binding tests
* The fortran solverdummy  now builds using a Makefile https://github.com/precice/precice/commit/7da1ff5b86b49ab0b3543fa2d2560d47080dd3b5  
This means SCons is **gone** from the systemtests
* The python solverdummies use the relocated config

All tests run!

fsimonis:
Regarding the remaining issues:

1) We add the user precice to the sudo group if sudo rights are required (installing into `/usr/local` or the package). 
2) Migrating the tests to user precice is easy to do in further PRs now that the base images work.
3) This looks good for the base images.

My suggestion: Lets merge this and continue on the remaining tasks.

MakisH:
>     * The fortran solverdummy  now builds using a Makefile [precice/precice@7da1ff5](https://github.com/precice/precice/commit/7da1ff5b86b49ab0b3543fa2d2560d47080dd3b5)
>       This means SCons is **gone** from the systemtests

I thought I committed it before the release, but it looks like [it's ignored](https://github.com/precice/precice/blob/7da1ff5b86b49ab0b3543fa2d2560d47080dd3b5/.gitignore#L22). :sweat_smile:

MakisH:
Shall we maybe rename `sudo` to `system` or `root` or `usr`? The term `sudo` sounds a bit irrelevant here and is inconsistent with the `home`.

fsimonis:
@MakisH you are right about `sudo`, however, we install it in `/usr/local/`, so `usr` and `root` are misleading. `system` is better, but that could still could be confused with the packages. What about `local`?

fsimonis:
The reason the system tests are failing is that they are based on the _ubuntu1604 docker images, which don't exist anymore.

So what base image should we use as our new defaut?
1) the DEB `package` install
2) the `sudo make install` into `/usr/local/`
3) the `home` install

MakisH:
> @MakisH you are right about `sudo`, however, we install it in `/usr/local/`, so `usr` and `root` are misleading. `system` is better, but that could still could be confused with the packages. What about `local`?

Then let's name it `usr_local` ;) (or `usrlocal`) Although I am not sure about the purpose of this, if we already have another prefix (`home`). I would drop it.
 
> So what base image should we use as our new defaut?
> 
>     1. the DEB `package` install
> 
>     2. the `sudo make install` into `/usr/local/`
> 
>     3. the `home` install

I would use the `package`, since it:
* is our recommended way for new users, so it will maybe affect the most
* tests more features (also the `make package`)
* we could then maybe use it to get automatically built packages

MakisH:
We could also test e.g. Ubuntu 18.04 with `package` and Ubuntu 16.04 with `home`.

MakisH:
Although I find it good to test everything possible once in a while, I think we should make sure to test a few combinations that make sense.

In this sense, I am not sure we should test a matrix that includes e.g. testing `home` for both Ubuntu 18.04 and 16.04 and then also `sudo` for both and `package` for both (actually, for 16.04, `package` does not make sense). Even though it looks nice testing six different combinations, the added value is small and we risk thinking that we test everything.

I would rather follow the following structure:
* stable dependencies (Ubuntu 18.04, variant `package`, with everything from APT) = see what most users do
* unstable / bleeding edge (Arch, latest versions) = see what will come in the future
* ancient (Ubuntu 16.04, variant `home`, with the oldest working versions) = make sure we did not raise our requirements, account for users that build some dependencies manually on clusters etc.

I believe that `sudo` and `home` do not test anything much different (as they both rely on prefix) and I would keep the `sudo` optional. The `home` is a stricter case.

In any case, the additional tests will be happening on a less frequent basis (e.g. weekly / monthly / per release).

shkodm (approved):
Looks good. I will adjust the Travis config and merge it later today. As discussed with @MakisH and @fsimonis I will leave only 16.04 home non-package and 18.04 package build as normal builds, and the rest we will test as cron jobs less frequently.

MakisH:
> Looks good. I will adjust the Travis config and merge it later today. As discussed with @MakisH and @fsimonis I will leave only 16.04 home non-package and 18.04 package build as normal builds, and the rest we will test as cron jobs less frequently.

I would also make the "16.04 home" less frequent for now, but this is not so important. Let's see how the total time goes.

shkodm:
All tests are [passing](https://travis-ci.org/precice/systemtests) on Travis now. I am merging. I will add tests for Ubuntu 18.04 in the subsequent PR.

Issue #36, opened 2019-02-25
Split the Docker base images into Dependencies + preCICE


MakisH:
It currently takes more than 30min to build preCICE on Travis, as it every time needs to build the dependencies. This is not really a problem, but can be annoying and can lead to unexpected external errors (e.g. if the Boost download link gets temporarily broken).

I think we could solve this by moving this part to another image, e.g. named `precice_depends` and only update it manually when needed.

- [ ] Split the Ubuntu 16.04 base image
- [ ] Split the Ubuntu 18.04 base image

https://github.com/precice/systemtests/blob/1475ddf5d3108931b0f392ce088e8598db7767d0/Dockerfile.Ubuntu1604#L6-L24

fsimonis:
I like this idea __a lot__!

Building the necessary subset of boost for each build is a waste of time and electricity. Especially as we pin the version to `1.65.1`.

BenjaminRodenberg:
Another possible use for this: Develop preCICE on any system you like. How?

1. build the image: `docker build -t myprecice:base -f Dockerfile .`
2. clone preCICE to a folder named `precice`
3. mount the `precice` folder on your host system in the container: `docker run -v /path/on/host/precice:/home/user/precice -ti myprecice:base`
4. build and run precice inside container. Develop on host.

I currently use this setup within Docker on windows. If we document this approach properly and provide the necessary Dockerfiles, this might be useful for some users, as well.

BenjaminRodenberg:
One thing that I would like to add to this issue:

* provide a non-root user + home directory (`USER precice`, `ENV HOME=/home/precice`)
* add `pip3` to the provided dependencies, since we (or more honest: I) use python a lot

BenjaminRodenberg:
Another possibly useful thing: [Already create the required `Output` folder](https://github.com/precice/systemtests/blob/bf39313461b899b49bc47b5e4ad443259d5bfe87/Test_bindings.Ubuntu1804/Dockerfile#L55-L56). This folder is currently needed to succesfully run the systemtests. If this (possibly empty) folder does not exist, the systemtests fail. Here, we should again think about reasonable permissions and provide an environmental variable `$OUTPUT`.

Of course, there are also other workarounds for this problem, like an option in our python scripts that deactivates output.

BenjaminRodenberg:
#145 is basically a duplicate of this issue (with some more details). I'm closing #145 now. Let's deal with the issue here.

PR #56, opened 2019-03-08
use sudo command instead of root user

BenjaminRodenberg:
Switching between users `root` and `precice` for executing commands that need `sudo` privileges (e.g. `dpkg` or `apt-get install`) is replaced by the user `precice` using the `sudo` command. This also reduces layer size.

If we want to be able to use the `sudo` command inside docker, this requires the following changes:

* `sudo` has to be installed via `apt-get`
* the sudoers file has to be modified, such that `precice` can call `sudo` without having to enter a password.

### DIFF
--- a/Dockerfile.Ubuntu1604.package
+++ b/Dockerfile.Ubuntu1604.package
@@ -34,8 +34,10 @@ RUN wget -nv 'https://dl.bintray.com/boostorg/release/1.65.1/source/boost_1_65_1
 # Rebuild image if force_rebuild after that command
 ARG CACHEBUST
 
-# create user precice
-RUN useradd -ms /bin/bash precice && adduser precice sudo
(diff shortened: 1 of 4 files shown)


fsimonis (approved):

BenjaminRodenberg:
What is the status of this PR? Do we still need it or does master already use cmake + proper users and rights?

BenjaminRodenberg:
https://github.com/precice/systemtests/issues/36#issuecomment-509626659 is related to this PR

shkodm:
This can be closed. We decided on another scheme, where we use root user for the base image dependency install, and maintain user `precice` (in group `precice`  with specified `uid` and `gid` ) for everything else, including running tests. I still need to update it in another base images, but `Ubuntu1604.home` on which most of tests run now follows this scheme.

PR #58, opened 2019-03-14
Add architecture overview

fsimonis:
This PR adds an overview of the system-tests architecture.
It aims to be more visible than an issue and more detailed in terms of responsibilities and restrictions.

The architecture doc includes:
* Layers
* Naming Scheme with Grammar
* Responsibilities
* Restrictions

This is based on the issue #14.
@BenjaminRueth @floli Please give feedback on the correctness of this doc.

### DIFF
--- a/doc/architecture.md
+++ b/doc/architecture.md
@@ -87,6 +87,16 @@ Base images will be matched against Tests in the following way:
         * drop a feature and repeat matching
       * else run Test with matching `TestClass` without `SystemSpec`
 
+Examples:
+
+Base Image                        | Matching Tests
+---                               | ---
+`Dockerfile.Ubuntu1804`           | `Test_of-of.Ubuntu1804`, _(no matching su2)_
+`Dockerfile.Ubuntu.1804.Boost160` | `Test_of-of.Ubuntu1804.Boost160`
+`Dockerfile.Ubuntu.1804.NoPETSc`  | `Test_of-of.Ubuntu1804`
+`Dockerfile.Ubuntu1604`           | `Test_of-of`, `Test_su2.Ubuntu1604`
+`Dockerfile.Ubuntu1604.Boost160`  | `Test_of-of`, `Test_su2.Ubuntu1604`
+
 ### Restrictions:
 * Run commands as user `precice`
 * Use `pip install --user` to install python dependencies
--- a/doc/architecture.md
+++ b/doc/architecture.md
@@ -1,0 +1,96 @@
+# Architecture
+
+The architecture of the system tests consist out of two layers.
+The interface between the layers are docker images.
+
+## Grammar
+
+```
+// terminals
+System         = string
+TestClass      = string
+Feature        = string
+Dockerfile     = "Dockerfile"
+TestPrefix     = "Test"
+.              = "."
+_              = "_"
+
+// Non-terminals
+SystemSpec     = System [. Feature]*
+Base           = Dockerfile . SystemSpec
+TestSpec       = TestClass | TestClass SystemSpec
+Test           = TestPrefix _ TestSpec
+```
+
+## Base Images
+
+### Naming Convention: 
+Described by Grammar: `Base`
+
+Examples:
+```
+Dockerfile.Ubuntu1804
+Dockerfile.Ubuntu1804.Boost160
+Dockerfile.Ubuntu1804.NoPETSc
+Dockerfile.Ubuntu1604
+Dockerfile.Ubuntu1604.Boost160
+```
+
+### Responsibilities:
+The image needs to provide:
+1) Software:
+    1) `cmake`
+    2) `make`
+    3) `git`
+    4) a C++ compiler
+    5) MPI binaries
+    6) `python3`
+    7) `python3-dev`
+    8) `pip`
+2) all dependencies necessary to build and use preCICE
+3) a user `precice` with a home directory containing:
+    * `src` a checkout of the git repository
+5) a discoverable install of preCICE:
+    * `pgk-config --exist precice` returns true
+    * `#include <precice/SolverInterface.hpp` has to work either out-of-the-box or by passing the flag `-I` using pkg-config.
+    * `ld -lprecice` has to work either out-of-the-box or by passing the flag `-L` using pkg-config.
+    * `binprecice` and `testprecice` have to be in `PATH`
+    * `PRECICE_ROOT` hast to be set so that at least `runprecice` will be able to run the tests.
+
+_Optional: after building preCICE, the "base" test-set has to be executed with `make test_base`!_
+
+## Tests
+
+### Naming Convention: 
+Described by Grammar: `Base`
+
+Examples:
+```
+Test_of-of
+Test_of-of.Ubuntu1804
+Test_of-of.Ubuntu1804.Boost160
+Test_su2.Ubuntu1604
+```
+
+### Matching Logic:
+
+Base images will be matched against Tests in the following way:
+* For each `Base` image:
+  * For each `TestClass`:
+    * For each `Test` with matching `TestClass` and existing `SystemSpec`:
+      * if `Test` and `Base` have matching `SystemSpec`
+        * run the test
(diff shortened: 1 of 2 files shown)


doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+      * if `Test` and `Base` have matching `SystemSpec`
+        * run the test
+      * else if `Base` `SystemSpec` has `Features`:
+        * drop a feature and repeat matching
+      * else run Test with matching `TestClass` without `SystemSpec`
+
+### Restrictions:
+* Run commands as user `precice`
+* Use `pip install --user` to install python dependencies
+* Avoid using `apt-get` or other system specify package managers
```
floli, line 103: Why that? That is the default way of installing software and in any case I would prefer it to using `pip`, `gem`, `cpan`, ... 

It's also the way the users probably install the software and provides more stable package versions.

I think it should be "use standard package manager unless the package is not available there".

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+* Do not use system specify package manager in a `Test` without `SystemSpec`
+
+### Responsibilities:
+
+* A failing `RUN` marks the test as failed.
```
floli: I am not really sure what is meant here. EDIT: Got it.

Suggestion: " A failing `RUN` command in the Dockerfile marks the test as failed."

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+    7) `python3-dev`
+    8) `pip`
+2) all dependencies necessary to build and use preCICE
+3) a user `precice` with a home directory containing:
+    * `src` a checkout of the git repository
+5) a discoverable install of preCICE:
+    * `pgk-config --exist precice` returns true
+    * `#include <precice/SolverInterface.hpp` has to work either out-of-the-box or by passing the flag `-I` using pkg-config.
+    * `ld -lprecice` has to work either out-of-the-box or by passing the flag `-L` using pkg-config.
+    * `binprecice` and `testprecice` have to be in `PATH`
```
floli: `$PATH`

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+    * `PRECICE_ROOT` hast to be set so that at least `runprecice` will be able to run the tests.
```
floli: `$PRECICE_ROOT`

What is `runprecice`?

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
+# Architecture
+
+The architecture of the system tests consist out of two layers.
```
floli: Wording seems to be weird `consists out of two`. Also, clarify what the layers are.

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+System         = string
+TestClass      = string
+Feature        = string
+Dockerfile     = "Dockerfile"
+TestPrefix     = "Test"
+.              = "."
+_              = "_"
+
+// Non-terminals
+SystemSpec     = System [. Feature]*
```
floli, line 22: So a `SystemSpec` can only have one feature. `Arch.Boost160` is ok, `Arch.Boost160.PETSc3.8` is not`. Do I understand that correctly?

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+    * `ld -lprecice` has to work either out-of-the-box or by passing the flag `-L` using pkg-config.
+    * `binprecice` and `testprecice` have to be in `PATH`
+    * `PRECICE_ROOT` hast to be set so that at least `runprecice` will be able to run the tests.
+
+_Optional: after building preCICE, the "base" test-set has to be executed with `make test_base`!_
+
+## Tests
+
+### Naming Convention: 
+Described by Grammar: `Base`
```
floli: Isn't that described by grammer `TestSpec`?

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+Dockerfile     = "Dockerfile"
+TestPrefix     = "Test"
+.              = "."
+_              = "_"
+
+// Non-terminals
+SystemSpec     = System [. Feature]*
+Base           = Dockerfile . SystemSpec
+TestSpec       = TestClass | TestClass SystemSpec
+Test           = TestPrefix _ TestSpec
```
floli, line 25: Should't this be:
```
TestSpec       = TestClass | TestClass.SystemSpec
Test           = TestPrefix_TestSpec
```
(removed white space, add period)

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+```
+Dockerfile.Ubuntu1804
+Dockerfile.Ubuntu1804.Boost160
+Dockerfile.Ubuntu1804.NoPETSc
+Dockerfile.Ubuntu1604
+Dockerfile.Ubuntu1604.Boost160
+```
+
+### Responsibilities:
+The image needs to provide:
```
floli: Use must or should here. I would say should, to make it less strict.

doc/architecture.md
```diff
@@ -0,0 +1,96 @@
...
+```
+Test_of-of
+Test_of-of.Ubuntu1804
+Test_of-of.Ubuntu1804.Boost160
+Test_su2.Ubuntu1604
+```
+
+### Matching Logic:
+
+Base images will be matched against Tests in the following way:
```
floli, line 80: Could you give some examples here?

floli (changes requested):

doc/architecture.md
fsimonis: The wildcard means 0 or more features

doc/architecture.md
fsimonis: Its a typo. It should be `testprecice`

doc/architecture.md
fsimonis: The whitespace is only in the grammar to separate different parts, it is not it the result. The examples should clarify this.

The period was missing. Thanks!

doc/architecture.md
fsimonis: It should say `Test`

doc/architecture.md
fsimonis: My intention is to put a `must` here.

My reasoning is that by making this strict, the system tests can rely on this being installed and can thus be more generic and lightweight.
What do you think?

doc/architecture.md
fsimonis: I'll use your examples from the issue, ok?

doc/architecture.md
fsimonis: This definitely needs some clarification.

Avoid meaning: "try to rely on the software installed in the base image"

Using the system package manager has a few back-draws:
1) It results in a non-generic/system-specific image
2) It requires administrative rights
3) It increases the size of the test image
4) Especially python packages are often (Ubuntu) hopelessly outdated. User install / virtualenv fixes this. Pip is the python package manager, so we should use that.

MakisH: Interestingly, pip [discourages using the distribution packages on Ubuntu/Debian](https://packaging.python.org/guides/installing-using-linux-tools/#debian-ubuntu):

> Warning: Recent Debian/Ubuntu versions have modified pip to use the “User Scheme” by default, which is a significant behavior change that can be surprising to some users.

I have also experienced bad situations with python/pip from the Ubuntu repositories.

floli: `pip` gives us rather unstable versions, as it always installs the newest, irrc. Yes, you can likey select a specific version.
Why does using `apt-get` or alike increases the size of the image more than `pip`?

Yes, If the Debian packages are outdated, we can and should fall back to pip. But i still think we should stick to the system packages if it works.

I kind of doubt that we will be able to get test images that are 100% independent from the base image, i.e., work for Debian and Arch.

fsimonis: `apt-get` ties the image to a certain system.

I think we should use `pip` as long as it allows us to avoid system specific images.

The bindings image is an example of an image that should be system-independent. One can install all python requirements via `pip'.

shkodm:
I close this, as discussed with @fsimonis

PR #60, opened 2019-03-21
Let systemtests run on PRs to the adapters

### LINKED ISSUE #50: Let adapters run systemtests based on pull request repo
Currently adapters, that use systemtests for their CI ( using https://github.com/precice/systemtests/pull/22 ) do not have a possibility to check pull requests. What is worse, pull request  will be built and run, but using the master branch of the adapter,  so it will pass all the time, even if it introduces breaking changes (for instance as can be seen in [ this pull request ]( https://github.com/precice/calculix-adapter/pull/14 ) ). 
    That is definitely not desired . Temporary solution is just to prohibit builds from pull request to not confuse everyone, however better solution would be to run systemtests using proper repository.

shkodm:
Basically it extends functionality of #22, so now we can also test different branches, and pull requests on the adapters. 
For every adapter one can decide what branches should bested, e.g. as done in [openfoam adapter travis config](https://github.com/precice/openfoam-adapter/blob/master/.travis.yml). 

Implementation is done a way of modifying Dockerfiles of systemtests of a specific adapter, so that after each `git clone $adapter` we also switch to a branch and fetch a pull request if Travis detects, that is triggered by it. That way one does not need to care about this functionality when adding new systemtests, as it will work automatically.

I tested on a several pull requests and branches of openfoam-adapter and it seems to work as expected.

Closes #50

### DIFF
--- a/trigger_systemtests.py
+++ b/trigger_systemtests.py
@@ -10,7 +10,6 @@ import os
 import time
 from sys import exit
 import argparse
-from string import Template
 from urllib.parse import urlencode
 from urllib.request import Request, urlopen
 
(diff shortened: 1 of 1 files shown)


trigger_systemtests.py
```diff
@@ -39,20 +38,55 @@ def get_json_response(url, **kwargs):
 
     return json_response
 
+def adjust_travis_script(script, user, adapter):
```
MakisH, line 41: Maybe rename `adapter` to `repository` for generality?

shkodm: Actually `adapter` here is just the name of the adapter (e.g. 'calculix'). I then query real repository name from `nm_repo_map`

PR #66, opened 2019-05-09
Use ccache during building of precice images

### LINKED ISSUE #52: Add ccache to base images
To add ccache in the system tests, we need to do the following:

1) Add a cached directory for ccache in Travis
2) Bind `$HOME/.ccache` form inside the docker base images to the cached directory in travis.
3) Install [ccache](https://ccache.samba.org/download.html) in the base images and add `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache` to the preCICE configuration.

shkodm:
This adds usage of ccache during building of preCICE images. There is no obvious way to bind the directory on travis to the directory in container during the build so that it can be accessed
( in contrast it is quite easy during running containers as can be see in [this blog post](http://frungy.org/docker/using-ccache-with-docker) ).
The way I went with it  `rsync`-ing directories created during image build and the one, that is cached by Travis, with subsequent removal of extra cache in the image. Everything can still be built without ccache by simply not specifying address from which to `rsync` in the docker `--build-arg`.

I maintain separate cache for different variants (16.04, 18.04), since I am not sure how otherwise things will get messed up. Everything seems to be working as tested on my fork. 

Local testing remains unchanged apart from addition of ccache.

Closes #52

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,3 @@
 *
+!post_install_hook.sh
+!pre_install_hook.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
   - docker
 cache:
   directories:
-    - $HOME/ccache
+    - $HOME/cache
 jobs:
   include:
 
@@ -26,7 +26,7 @@ jobs:
     if: fork = false
     name: "Ubuntu 16.04 home"
     script:
-    - source setup_ccache_sync.sh ubuntu1604
+    - source setup_docker_ci_cache.sh ubuntu1604
     - docker build --build-arg CCACHE_REMOTE=$CCACHE_REMOTE -f Dockerfile.Ubuntu1604.home -t precice .
     after_success:
     - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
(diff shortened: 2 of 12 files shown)


fsimonis (commented):
Wouldn't it be simpler to use straightforward volume mounting?
`-v /ccache/$DISTRO/:~/.ccache`
Then you can simply cache the entire `/ccache/` directory.

The upsides are:
1) It requires not changes in the base images (They can just use ccache all the time)
2) It requires no additional scripts or a running rsync daemon

shkodm:
@fsimonis , No, I would have gone with the way you proposed if this was possible. But for whatever reason it is only possible to use `-v /ccache/$DISTRO/:~/.ccache` during `docker run`, not `docker build`, see for instance [this stackoverflow question](https://stackoverflow.com/questions/26050899/how-to-mount-host-volumes-into-docker-containers-in-dockerfile-during-build)

fsimonis:
I see, your proposal is the way to go then.

I have a few additional comments:
* You could refactor the `if[ ! .z CCACHE_REMOTE]; then`-blocks out into separate scripts e.g. `pre-compilation-hook` and `post-compilation-hook` and then copy them into the docker images using `ADD`. This keeps the docker commands clean, makes the scripts easier to maintain and gives you the space to properly document what is going on.
* You should clean `~/.ccache` unconditionally as you always use `ccache`.
* The `pre`- and `post-compilation-hooks` should be part of the same `RUN` command and layer.

shkodm:
@fsimonis, that was a good idea indeed. I refactored and generalized the code into `pre_install_hook.sh`h and `post_install_hook.sh`, that can be used with for any suitable dependency with `--ccache` flag for ccache caching and additionally could cache downloads and source code of other dependencies. I will describe more in another PR, where I will use it to add caching for dependencies.

shkodm:
@fsimonis, any other suggestions? Otherwise I am merging.

shkodm:
Upon some discussion and iterating on ideas, me and @fsimonis decided on the other way to go regarding the whole caching situation. I close this for now.

PR #72, opened 2019-06-22
Switch Arch to cmake

### LINKED ISSUE #37: Build preCICE images with CMake instead of SCons
We currently build our base preCICE images with SCons:

https://github.com/precice/systemtests/blob/1475ddf5d3108931b0f392ce088e8598db7767d0/Dockerfile.Ubuntu1604#L46

We should [build these with CMake](https://github.com/precice/precice/wiki/Building:-Using-Cmake).

- [x] Build with CMake on Ubuntu 16.04
- [x] Build with CMake on Ubuntu 18.04

We could still have one target to build with SCons for a while, but we should treat CMake as our default.

shkodm:
The installation basically mirrors the installation we have for all the the other base images, apart from the dependency part. 

Closes #37.

### DIFF
--- a/Dockerfile.Arch
+++ b/Dockerfile.Arch
@@ -29,6 +29,8 @@ ARG branch=develop
 RUN git clone --branch $branch https://github.com/precice/precice.git /home/precice/precice
 WORKDIR /home/precice/precice
 
+# add oversubscribe to fix travis
+ENV mpiexec='mpiexec -oversubscribe'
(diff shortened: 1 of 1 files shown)


floli (approved):
:+1:

shkodm:
I added additional `oversubscribe` flag to `mpiexec` to handle tests on Travis, that were failing.

PR #75, opened 2019-07-04
Use mpich and test mpi ports 

### LINKED ISSUE #73: Test MPICH + MPIPorts
Tests for the MPIPorts are missing in the systemtests. We only run `make test_base` in our current tests. This means, we should add a `Dockerfile.UbuntuXXXX.home.MPICH` where `MPICH` is used as mpi implementation; this allows us to run `make test` (which includes the tests for MPIPorts; see [here](https://github.com/precice/precice/wiki/Tests#known-mpi-port-test-failures))

BenjaminRodenberg:
Closes #73 

I added a new Docker base image that uses mpich instead of OpenMPI. This allows running the tests that are expected to fail if we use OpenMPI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ jobs:
     - docker push $DOCKER_USERNAME/precice-ubuntu1604.home-develop:latest
 
   - stage: Building preCICE
-    if: type = cron
+    if: fork = false
     name: "Ubuntu 16.04.sudo"
     script:
     - docker build -f Dockerfile.Ubuntu1604.sudo -t precice .
@@ -41,7 +41,7 @@ jobs:
     - docker push $DOCKER_USERNAME/precice-ubuntu1604.sudo-develop:latest
 
   - stage: Building preCICE
-    if: type = cron
+    if: fork = false
     name: "Ubuntu 16.04.package"
     script:
     - docker build -f Dockerfile.Ubuntu1604.package -t precice .
@@ -53,7 +53,7 @@ jobs:
 
   - stage: Building preCICE
     name: "Ubuntu 18.04.home"
-    if: type = cron
+    if: fork = false
     script:
     - docker build -f Dockerfile.Ubuntu1804.home -t precice .
     after_success:
@@ -64,7 +64,7 @@ jobs:
 
   - stage: Building preCICE
     name: "Ubuntu 18.04.sudo"
-    if: type = cron
+    if: fork = false
     script:
     - docker build -f Dockerfile.Ubuntu1804.sudo -t precice .
     after_success:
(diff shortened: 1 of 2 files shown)


BenjaminRodenberg:
@shkodm at some points `type = cron` is used, sometimes `fork = false` is used in `.travis.yml`. I was not totally sure about the purpose of these arguments and therefore I just set `fork = false` everywhere. Can you please check on this?

.travis.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     - docker push $DOCKER_USERNAME/precice-ubuntu1604.home-develop:latest
 
   - stage: Building preCICE
-    if: type = cron
+    if: fork = false
```
shkodm: Change it back to `'type = cron`

.travis.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     - docker push $DOCKER_USERNAME/precice-ubuntu1604.sudo-develop:latest
 
   - stage: Building preCICE
-    if: type = cron
+    if: fork = false
```
shkodm: Same

.travis.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
   - stage: Building preCICE
     name: "Ubuntu 18.04.home"
-    if: type = cron
+    if: fork = false
```
shkodm: Same

.travis.yml
```diff
@@ -64,7 +64,7 @@ jobs:
 
   - stage: Building preCICE
     name: "Ubuntu 18.04.sudo"
-    if: type = cron
+    if: fork = false
```
shkodm: Same

shkodm (changes requested):
`fork = false` just means we don't even want to build and try push preCICE on fork, because it will fail to push to the dockerhub due to access rights. 
`type = cron` means we run as a cron job, nightly (obviously this also implies that tests probably would not run on forks, unless you configure cron jobs in forked repo ). As far as I remember, the conclusion of discussion with @MakisH at some point was that we add most tests to run nightly, apart from the essentials.   So you should probably change everything here to `type = cron`

.travis.yml
```diff
@@ -73,6 +73,17 @@ jobs:
...
     - docker push $DOCKER_USERNAME/precice-ubuntu1804.sudo-develop:latest
 
+  - stage: Building preCICE
+    name: "Ubuntu 18.04.sudo.mpich"
+    if: fork = false
+    script:
+    - docker build -f Dockerfile.Ubuntu1804.sudo.mpich -t precice .
+    after_success:
+    - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+    - docker images
```
shkodm: You can reduce building and tagging to one command, as fixed here #76

BenjaminRodenberg: 7fa22efc1ca465ac6a7c0e590874382ac5b3f2df fixes this

shkodm (approved):

PR #79, opened 2019-07-12
Add test case for FEniCS

BenjaminRodenberg:
Adds the [FEniCS-FEniCS tutorial](https://github.com/precice/tutorials/tree/master/HT/partitioned-heat/fenics-fenics) to our tests. There is an assertion in the tutorial that assures that the numerical solution is close to the analytical solution. Additionally the iterations and convergence in the coupling is checked against `Test_fe-fe.Ubuntu1804/referenceOutput`.

~~Possible Extension: The FEniCS-OpenFOAM test case [flowOverPlate](https://github.com/precice/tutorials/tree/master/CHT/flow-over-plate/buoyantPimpleFoam-fenics). Here we would need the usual setup with reference results etc.~~

Unfortunately I was only able to get FEniCS running under Ubuntu 18.04. Not under Ubuntu 16.04.


Test_fe-fe.Ubuntu1804/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+# Downloading tutorial and run it
+WORKDIR /
+RUN git clone https://github.com/precice/tutorials.git
+WORKDIR /tutorials/HT/partitioned-heat/fenics-fenics
+RUN pip3 install --user -r requirements.txt && \    
+    python3 heat.py -d & python3 heat.py -n
+
+# Check output
+
+## TODO: compare iteration count to reference result!
```
shkodm: Why did not you add it now?

shkodm (approved):

Test_fe-fe.Ubuntu1804/Dockerfile
BenjaminRodenberg: Good point. I will add some reference results for the QN iterations. 

The quality of the solution is already checked by [`assert (error_total < total_error_tol)`](https://github.com/precice/tutorials/blob/1f44eb812d4d6bf4df489ad27f8d6dd3be1db5ac/HT/partitioned-heat/fenics-fenics/errorcomputation.py#L9). Therefore, we do not need an additional check.

shkodm: @BenjaminRueth , what is the status here? Should I merge it, or are going to add reference for the QN iterations?

BenjaminRodenberg: Thanks for the reminder. I totally forgot about this issue. I will take care of it next week and then we can merge it.

BenjaminRodenberg:
If https://github.com/precice/systemtests/pull/79/commits/fc9e4aa658e4d3da203b5e1734fb2def60ddacc7 fails and https://github.com/precice/systemtests/pull/79/commits/71fad1731d1ad15a2285cf9fa16ee3ff04dd4e6d succeeds, we can `git revert fc9e4aa658e4d3da203b5e1734fb2def60ddacc7` and merge this PR.

PR #80, opened 2019-07-26
Split tests in several stages and switch to docker-compose

### LINKED ISSUE #55: Review access rights of our docker containers
As already mentioned in #54 our current practice of using `root` for our system tests is problematic from a technical point of view (mpicexec, pip...). Additionally, it is also problematic with respect to security and portability aspects of our docker images (see for example [here](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b)). 

We should use a unprivileged user (e.g. `precice`) in all our tests instead of `root`. 

*A few open questions:* The name `precice` is not ideal, since we already have many things called `precice` in our project (this name is used for example in dcb07b6a59c39dc6d56750faf6aefb2ba50c69d4). Having a user called `precice` on top does not really help. 

* Any suggestions for a nice name?
* Is there a way to make this username configurable from outside/from a central point, such that we do not have to change all our Dockerfiles (again), if we change our mind?

### LINKED ISSUE #62: Decrease Travis output
Currently, the screen output we get on Travis is so long, that it does not even fit in the normal Travis viewer. Even further, it's difficult to find out what the problem is.

We need ways to reduce the output, without losing much information. We could, e.g. redirect most output to files that would go into the `precice_st_output` repository.

We already have e.g. [this script](https://github.com/precice/systemtests/blob/master/Test_of-ccx.Ubuntu1604/reducing_log_of-ccx) reducing the simulation output.

shkodm:
This introduces splitting of systemtests into proper stages and dockerfiles. 
@MakisH, @fsimonis and me thought that such architecture is way cleaner than the previous one.

The new architecture is outlined below: 

__0. **Building solvers and their dependencies**__

**See**: `Dockerfile.solver_name`
Not executed on every build, since solvers do not change usually. Done for CalculiX adapter ( build of ARPACK & spooles ),  openFOAM adapter (full build), SU2 adapter (just .ccache, since the adapter changes some parts of the solver), deal.ii (full build)

__1. **Building preCICE**__

Nothing changes here, apart from introduction of defined `uid` and `gid` for the preCICE
user. 

__2. **Building adapters**__

**See**: `Dockerfile.adapter-name`
Done using preCICE as the base image, and importing needed data from the solver images, that were created in the step 0.
Everywhere adapters are installed under `/home/precice/adapter-name` for consistency. There are additional folders, created for data exchange, input and output `/home/precice/Data/Input`, `/home/precice/Data/Output`, `/home/precice/Data/Exchange`.

__3. **Running adapters**__ using [docker-compose](https://docs.docker.com/compose/).

**See**: `TestCompose_case-name/docker-compose.yml`
This is a step, where the actual tests are run.  All compose-based tests now have prefix `TestCompose_`. I kept bindings following the old running strategy. So docker-compose is a prerequisite now. Roughly following is executed: 

  - Build of an image `tutorial_data`, based on the alpine image, that clones the original `tutorials` repository, does some `sed`-ing to replace parts in the preCICE config, copies needed config files (e.g `precice-config.xml`) into `config` folder. Everything is done using `precice` user again. 
  - Creating named volumes for the input data for every adapter, output, and configuration files, using data created by `tutorial_data` image. Additionally volume for the data exchange is created.
  - Starting the adapters, that are needed for the tests, mapping of created named volumes to `Input`, `Output`, `Exchange` folder in every adapter. 
  - Copying back the data from the `Output` volume to the host and comparing it with the reference
  - Deleting all images, containers, shutting down networks, full cleanup

Additional niceties - output is displayed only for the participants, that fail, so in general it will be mostly clean (this automatically closes #62 ). This is accomplished using `silent_compose.sh` script. There is also nice color separation between the output from the different solvers. If participant fails, I keep the containers, volumes, etc for further inspection. All can be removed with `docker-compose down -v` in the folder that is related to the test case. 
Also I now maintain `precice` user and `precice` group across all the images, so this will close #55.

Nothing changes in terms of the interfaces, everything, including triggering builds on the adapters, of clean rebuilds should continue working. Both local tests and runs on Travis are working, as far as I tested. Deciding which image to pull, etc is done using environment variables in the `docker-compose.yml` file.


MakisH:
I have not seen these changes in detail (I would happily give a detailed review when this PR ready), but some quick observations/questions/comments:

* Why are the Dockerfiles based on Ubuntu 16.04? Are we planning on moving to 18.04 in a separate PR? 16.04 reaches end-of-life next year, while 18.04 will be maintained until 2028.
* Why do we need e.g. `COPY --from=openfoam /opt/openfoam4 /opt/openfoam4/`? Isn't this readily available from the base image?
* > I kept bindings following the old running strategy.
  
   Is there a technical issue or just postponing this for now?

Overall, this looks quite better than before! :D

shkodm:
@MakisH, 

> Why are the Dockerfiles based on Ubuntu 16.04? Are we planning on moving to 18.04 

Good catch, for the solver dockerfiles, I simply forgot to provide base image as the parameter. Once this is done, everything else should be run as before ( you can specify 18.04 as base and everything should run ). So it is simply switchable to 18.04 and can be done in a separate PR. 

> Is there a technical issue or just postponing this for now?

Well, not really technical issue, but bindings are lightweight, the tests are simple, so it is easier  to manage them when they are grouped together in one dockerfile. 

> Why do we need e.g. COPY --from=openfoam /opt/openfoam4 /opt/openfoam4/? Isn't this readily available from the base image? 

We have two stages (two bases), one preCICE-based, on the top of which we build the adapter-image, and the one where we import the clean openfoam4 build ( see https://docs.docker.com/develop/develop-images/multistage-build/ ). I could also do `COPY` directly without specifying the first stage, but then it is less flexible with respect to the `from` name.

shkodm:
@MakisH, it is ready. Feel free to review, I will be merging it soon.

fsimonis:
@shkodm I would prefer for the system tests to run clean by default and always fully tidy up.
You could introduce a `--keep` or `--keep-on-error` flag to the scripts to prevent them from cleaning up the failed systemtest.

shkodm:
I added some cleanup and better error handling. As far, as I tested everything is working. I am merging.

Issue #19, opened 2018-12-12
Comparison between results of different solver/adapter versions


shkodm:
During fixing test cases for Ubuntu 18.04 and bringing it to travis as well, I have noticed that some of the results differ. (See #17 for documentation ). Which lead me to think about more general question:  

-  How do we want to compare results for a different versions of solvers, adapters, and operating system? Right now, very simple comparison between files is done, so if some small numerical disturbances occur, systemtest will fail and produce almost completely useless output message. For example  for  `openFOAM <-> openFOAM `, when changing openFOAM from version 4 to version 5,  test case fails and produce  93 files, that one would need to compare manually with the reference to find out what is the reason of failure...With SU2-Calculix number of files is larger than a 1000.  
     
    In the end if one would go through  the files, he would find that difference between files is just the header of the output file (where openFOAM version is specified) and non-significant fluctuations in the floating point numbers.  So probably this test should not have failed in the first place. 
    
    My idea ( already discussed with @MakisH and @BenjaminRueth  ), is to have some kind of  "maximum numerical difference" and instead of  dumb comparison between folders with reference and actual results, compare only relevant numerical values between folders and make the test fail only if the difference is large enough. I already played around with it and have the script the does the generic parsing and computing "average relative numerical difference" between folders  [here](https://github.com/shkodm/systemtests/blob/master/compare_results.sh). What else can we do? Of course, we can also just upload new results for each solver and adopter version, but  then the size of the repository will be growing quite large.  Any ideas or suggestions? @floli @MakisH @BenjaminRueth

MakisH:
If I understand correctly, the reference results in this case were produced with a different OpenFOAM version, compiled and ran on a different machine. Therefore, I expect them to be slightly different and, after we evaluate the importance of the difference, update the reference results we will use from now on.

Overall, we should always check for the same version of everything (the oldest one we support, I would say).

The "non-significant fluctuations in the floating point numbers" is a bit tricky. If we repeat the same simulation in the same environment, we should receive identical results in our cases. A contributor should know that her changes affected the behavior (even by one digit) and she should evaluate the significance and investigate the reasons.

Apart from this, I like the idea of having a measure of the maximum difference we introduce (could be an error, could be an improvement).

Could you maybe point to some example data for your script?

shkodm:
Yea, you can test it with for instance: 
```bash 
git clone https://github.com/shkodm/systemtests 
git clone https://github.com/shkodm/precice_st_output 
```
[...]

PR #83, opened 2019-08-12
Add nutils-openFOAM test

### LINKED ISSUE #74: Add Nutils CHT tutorial as system test
Since a few weeks, there is also a tutorial with Nutils. 
https://github.com/precice/tutorials/tree/master/CHT/flow-over-plate/buoyantPimpleFoam-nutils
Would be great to also have this one as a system test.

Please note: There is no such thing as a Nutils adapter. Nutils is a FEM framework just as FEniCS, but we directly call the preCICE API from the Nutils application script.

shkodm:
This adds `Nutils-openFOAM` test.
 In principle  it is similar to all the tests rewritten with #80 (two docker compose services, one for openFOAM, one for Nutils, both have separate folders for input and output). The only difference is that here, since there is no adapter for Nutils, I don't add it to `Building the adapters` stage, but rather add separate dockerfile for its dependencies, from which the image is built with each run. 

Closed #74


shkodm:
As can be seen from  [this log](https://api.travis-ci.org/v3/job/571418982/log.txt) there is some slight floating point difference between results, that I obtain on my machine and results on Travis...I am not sure why. This is related to #19, but there were at least different Ubuntu version, here they are identical, so I am not sure what can be the problem, maybe related to hardware FP optimizations?

PR #85, opened 2019-08-13
Add openFOAM-CalculiX test 

### LINKED ISSUE #32: Add test for FSI with OpenFOAM-CalculiX
We need to add our [latest tutorial](https://github.com/precice/tutorials/tree/master/FSI/flap_perp/OpenFOAM-CalculiX) (case `flap_perp`) as a test case. This could need some tweaking of the current naming.

When porting it here, it would be good to reduce the simulated time, so that the test finishes quicker. Let's maybe add the parallel version (or both the serial and parallel), to cover more cases.

shkodm:
Closes #32 

This also adds build of preCICE with PETSc ( currently failing with PETSc below 3.7, so built procedure is slightly adopted ) and corresponding adapters, as well as fixes few bugs related to running with different base images.


shkodm:
This is currently failing due to problem when preCICE is built with PETSc. I merge this, since this is unrelated to the tests itself.See https://github.com/precice/precice/issues/499.

PR #87, opened 2019-08-20
Add documentation

shkodm:
This add brief documentation of the general architecture as well as some tips on development.


docs/development.md
```diff
@@ -0,0 +1,58 @@
...
+If something fails during running `docker-compose` due to communication problems or invalid input, the easiest way to debug is:
+
+1. Add sleep timer (e.g. `sleep 1000`) after running failed command
+2. Rerun the failed setup
+3. Attach to the failed container with e.g.
+```
+   docker-compose run -it calculix-adapter
+```
+4. You can then try to rerun command in `docker-compose.yml` and check the folders `/home/precice/Data/Input`, `/home/precice/Data/Exchange`, 
+`/home/precice/Data/Output`. Make sure the `precice-config.xml` is properly setup, input can be read and folders have proper rights.
```
BenjaminRodenberg: What does "proper rights" mean? I think some comments on this might be helpful.

From my perspective there are two points, where things with rights can go wrong:

1. **provide data to the container:** User `precice` needs the rights to read (and write?) the data in the mounted folder.
2. **obtain data from the container** The host (or another container) needs the rights to read (and write?) the data written by the container into the folder.

My current use case is the following: I `docker run -ti` two containers (e.g. `fenics-adapter`) that do all the computation. The host provides the case setup (e.g. `/tutorials/HT/partitioned-heat/fenics-fenics`) using `-v`. If the host does not set the user rights properly, `precice` cannot access the case and write output to the folder and aborts.

BenjaminRodenberg (commented):
I think it would be helpful to provide instructions about the following usecase:

* execute code in docker containers (one for each participant)
* provide setup and do visualization/postprocessing on host

This is useful, if you cannot/don't want to install the software necessary for running a coupled simulation on the host system. The use-case is not related to testing, but relies on the provided testing infrastructure and I think we can migrate the documentation to a place that fits better at a later point in time.

docs/development.md
shkodm: You can achieve it something like:

```
docker build -f Dockerfile.Ubuntu1804.home -t precice-user --build-arg uid=$UID --build-arg gid=$GID . 
docker build -f Dockerfile.fenics-adapter -t fenics-adapter-user --build-arg from=precice-user
```
where GID and UID are you group id (obtain e.g. with `id -a` on Linux). Then you run the container with:
```
docker run -ti --user=precice fenics-adapter-user /bin/bash
```

I you want to run two containers: 

```
docker volume create exchange
docker run --user=precice -it -v $(pwd)/HT/partitioned-heat/fenics-fenics:/home/precice/Data/Input -v exchange:/home/precice/Data/Exchange fenics-fenics-user /bin/bash```
You need to adjust the `precice-config file`, similarly as I do for tests though

**Comment by BenjaminRodenberg** [comment]
@IshaanDesai is going to write a Dockerfile using CentOS for building preCICE and he is going to use the documentation.

File: docs/development.md (hunk 1 above)
Comment (BenjaminRodenberg) [reply docs/development.md]: Thanks! Did you already add these explanations to the documentation? As far as I understand `--build-arg uid` and `--build-arg gid` are not provided by docker, but you added them to the Dockerfiles (e.g. [here](https://github.com/precice/systemtests/blob/06fe89a4cd5af34ba8401b5c2f8631c6b1bd185a/Dockerfile.Ubuntu1604.home#L33-L34)) in order to sort out the user rights for our specific setup of docker containers.

Some questions the documentation should answer in this context:

* Why do we actually have to worry about creating a preCICE user?  
* Why not just run everything as `root` (which is often done in docker)? We, for example, need a non-root user for testing `make install` in a user folder. We rely at some places on `pip` for installing packages and you should not run `pip` as sudo (correct @fsimonis?). 
* What can we do with the `uid` and `gid` argument? See example above.

File: README.md [hunk 2]
```diff
@@ -1,79 +1,22 @@
...
-
-* preCICE Library path: `-L/home/precice/precice-install/libs` or `pkg-config --libs-only-L libprecice`
-* preCICE Include path: `-I/home/precice/precice-install/include` or `pkg-config --cflags-only-I libprecice`
-* preCICE library flag: `-lprecice` or `pkg-config --libs-only-l libprecice`
-
-# Adding CI to a new adapter
-
-As described in [#22](https://github.com/precice/systemtests/pull/22),  using `trigger_systemtests.py` it is possible to trigger custom systemtests from the travis job of the another adapter, therefore providing continous integration for the adapter on each commit to a particular branch. If you want to add another adapter, or modify tests that are run for it, you need to modify `nm_repo_map` and `nm_test_map` variables in the `trigger_systemtests.py`, that describe mapping between adapter name and repository and adapter name and set of test cases correspondingly.
+```
+   python3 local_test.py -s of-of su2-ccx -f Dockefile.Ubuntu1604.home
```
MakisH: Why do we have two scenarios/system tests in this example (`of-of` and `su2-ccs`)? Why does the help message show two arguments, one required and one optional? I guess this means "multiple tests can be provided in the same command", but this was a bit confusing in the beginning. A short explanation of these arguments here would be great.

MakisH: Is the `-f Dockefile.Ubuntu1604.home` correct? If I run it like this I just get the help message. (`-f` is for Docker itself, right?)

From the help message I understand that it should be `-d Dockefile.Ubuntu1604.home`, but then I get an error message:
```
$ python3 local_test.py -s of-of -d Dockefile.Ubuntu1604.home
  File "local_test.py", line 39, in <module>
    tests = filter_tests(tests, args.dockerfile)
  File "/home/makish/Desktop/systemtests/common.py", line 116, in filter_tests
    base_features.remove('Dockerfile')  # remove Dockerfile
ValueError: list.remove(x): x not in list
```

What am I doing wrong?

README.md
```diff
@@ -1,79 +1,22 @@
...
+``` 
 
-The adapter's Travis build would be considered successful if systemtests Travis build exits with status 'Failed' or 'Errored'. It would exit with 'Success', if the systemtests Travis build exits with 'Success' or 'Canceled'.
+More information can be found in the [docs](`www.github.com/precice/systemtests/docs`)
```
MakisH: Running the tests locally is indeed the most useful use case. However, I think it would be good to already add a few lines on what else this repository contains and a top-level view of what scripts are contained here. We could then point to the more detailed instructions for each use case.

In other words, there is a lot of work that is currently not shown directly in this Readme.

MakisH: This link is broken. Suggestion for relative path:
```suggestion
More information can be found in the [docs](docs/)
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
+# General architecture
```
MakisH: Naming this file `README.md` would make it shown by default on GitHub when one is at the `docs/` directory. Currently it needs additional thought to select the right one and it will become more difficult as the project grows.

docs/main.md
```diff
@@ -0,0 +1,61 @@
+# General architecture 
+
+Tests are executed in several stages: 
+
+## Building preCICE
+
+Several images of preCICE are built, based on different system (Ubuntu 16.04/18.04, Arch Linux) and different specifications
```
MakisH:
```suggestion
Several images of preCICE are built, based on different systems (Ubuntu 16.04/18.04, Arch Linux) and different specifications
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+of properties of the image ( root/non-root, package, different MPI versions ). This corresponds to the `Dockerfile.$systemname$.${system_spec}`
```
MakisH: Where are these properties defined? I know what `package` means in this context, but I am not sure if a new developer would immediately understand it.

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+Resultant images are then pushed to the dockerhub.
```
MakisH: A link would be useful:
```suggestion
Resultant images are then pushed to [Dockerhub](https://hub.docker.com/u/precice).
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Building solvers 
+
+Is done once for every solver, using `Dockefile.${solver_name}`. Resultant images are then stored on the dockerhub.
```
MakisH: A link would be useful:
```suggestion
Is done once for every solver, using `Dockefile.${solver_name}`. Resultant images are then pushed to [Dockerhub](https://hub.docker.com/u/precice).
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Building adapters
+
+To build the adapter we follow two-stage builds, first importing from the preCICE image then from the solvers image (if the solvers image is
```
MakisH: I think it reads better like this:
```suggestion
To build an adapter we follow a two-stage build, first importing from the preCICE image and then importing from the solver's image (if the solver's image is
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+actually relevant for building the adapter). We can import from different preCICE base images by specifying `from` argument during building of the
+the image, e.g:
```
MakisH: Duplicate
```suggestion
image, e.g:
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+```
+docker build --build-arg from=precice/precice-ubuntu1804.home-develop -f Dockerfile.openfoam-adapter -t openfoam-adapter-ubuntu1804.home-develop .,
```
MakisH: Typo?
```suggestion
docker build --build-arg from=precice/precice-ubuntu1804.home-develop -f Dockerfile.openfoam-adapter -t openfoam-adapter-ubuntu1804.home-develop .
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+```
+Note, that due to the fact that different adapter 'adapts' differently, the general procedure is slightly
```
MakisH:
```suggestion
Note, that due to the fact that each adapter "adapts" differently, the general procedure is slightly
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+different and is outlined below: 
+
+- **SU2 adapter**
+
+  Since SU2 adapter modifies SU2 source code, it makes no sense to copy clean SU2 build, because it will need to be rebuilt. Instead we just build the adapter
```
MakisH:
```suggestion
  Since the SU2 adapter modifies the SU2 source code, it makes no sense to copy a clean SU2 build, because it will need to be rebuilt. Instead, we directly build the adapted
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+  from  SU2 source after applying the necessary modifications.
```
MakisH: Together with the previous suggestion
```suggestion
  SU2 source.
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+- **dealii adapter**
```
MakisH:
```suggestion
- **deal.II adapter**
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+  Since dealii is a library and the adapter just needs to link with it, here we actually perform the import of dealii image and only build
```
MakisH:
```suggestion
  Since deal.II is a library and the adapter just needs to link with it, here we import the previously-built deal.II image and only build
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+  the actual adapter
```
MakisH:
```suggestion
  the adapter
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+actually relevant for building the adapter). We can import from different preCICE base images by specifying `from` argument during building of the
+the image, e.g:
+```
+docker build --build-arg from=precice/precice-ubuntu1804.home-develop -f Dockerfile.openfoam-adapter -t openfoam-adapter-ubuntu1804.home-develop .,
+```
+Note, that due to the fact that different adapter 'adapts' differently, the general procedure is slightly
+different and is outlined below: 
+
+- **SU2 adapter**
+
```
MakisH: For consistency with the rest:
```suggestion

```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+  Since SU2 adapter modifies SU2 source code, it makes no sense to copy clean SU2 build, because it will need to be rebuilt. Instead we just build the adapter
+  from  SU2 source after applying the necessary modifications.
+
+- **dealii adapter**
+  Since dealii is a library and the adapter just needs to link with it, here we actually perform the import of dealii image and only build
+  the actual adapter
+
+- **CalculiX adapter**
+  We only import the necessary libraries (Spooles, ARPACK), since CalculiX adapter modifies CalculiX source.
```
MakisH:
```suggestion
  We only import the necessary libraries (Spooles, ARPACK), since the CalculiX adapter modifies the CalculiX source code.
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+- **openFOAM adapter**
```
MakisH:
```suggestion
- **OpenFOAM adapter**
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+  Same as for deal.ii adapter.
```
MakisH:
```suggestion
  Same as for the deal.II adapter.
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+- **Nutils adapter**
+  There is no Nutils adapter. There is python script that communicates with Nutils and preCICE. We therefore provide Dockefile for it just for the where it is used,
```
MakisH:
```suggestion
  The Nutils adapter is a Python script that communicates with Nutils and preCICE. We therefore only provide a Nutils adapter Dockefile for the test running stage (not for the building stage).
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+  and don't build during adapter building stage, but during test running stage.
```
MakisH: Together with the previous suggestion:
```suggestion
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+- **FEniCS adapter**
+  We install FEniCS, together with adapter using `Dockefile.fenics-adapter` file.
```
MakisH:
```suggestion
  We install FEniCS, together with the FEniCS adapter, using the `Dockefile.fenics-adapter` file.
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+  
+
+All adapters are built using user `precice` with `gid` and `uid` equal to 1000 and of `precice` group. As a last build stage they create folder for the volume mapping in
```
MakisH:
```suggestion
All adapters are built using user `precice` with `gid` and `uid` equal to 1000 and of `precice` group. As a last building stage they create folder for the file system volume mapping in the
```
I added the `file system` because `volume mapping` could mean a different thing in preCICE.

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+- **FEniCS adapter**
+  We install FEniCS, together with adapter using `Dockefile.fenics-adapter` file.
+  
+
+All adapters are built using user `precice` with `gid` and `uid` equal to 1000 and of `precice` group. As a last build stage they create folder for the volume mapping in 
+`/home/precice/Data` folder.
+
+## Running tests
+
+At this stage, we spawn containers with the necessary solvers for the simulation. The execution is controlled using [docker-compose](https://docs.docker.com/compose/). Separate volumes are created for the input and the output of every adapter. Additional docker network is created and used for the communication between containers. Exchange of the data is done using another common volume. The general pattern across of the tests for such volumes to have locations `/home/precice/Data/Exchange`, `/home/precice/Data/Input`,
```
MakisH: Minor language suggestions:
```suggestion
At this stage, we spawn containers with the necessary solvers for the simulation. The execution is controlled using [docker-compose](https://docs.docker.com/compose/). Separate volumes are created for the input and the output of every adapter. An additional docker network is created and used for the communication between containers. Exchange of data is done using an additional, common volume. The general pattern across the tests for such volumes are paths such as `/home/precice/Data/Exchange`, `/home/precice/Data/Input`,
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+`/home/precice/Data/Output`, as explained above.
+
+The tests are based on the corresponding tutorials from the [tutorials repository](https://github.com/precice/tutorials). Several modifications are necessary to decouple the input to solvers, adjust simulation time, or change solver version and to modify `precice-config.xml` for simulation with docker. This is done with `Dockerfile.tutorial_data` image, that clones the tutorial data, adjusts and splits the input to different solvers. It is then is used as first container spawned by `docker-compose`.
```
MakisH: What do you mean by "to decouple the input to solvers"?

MakisH: Minor language suggestions:
```suggestion
The tests are based on the corresponding tutorials from the [tutorials repository](https://github.com/precice/tutorials). Several modifications are necessary to decouple the input to solvers, to adjust the simulated time, or to potentially change a solver's version, as well as to modify the `precice-config.xml` for the communication pattern we use in docker. This is done using the `Dockerfile.tutorial_data` image, which clones the tutorial data and adjusts and splits the input to different solvers. It is then used as the first container spawned by `docker-compose`.
```

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+Afterwards, when the simulation finishes the output is copied from all the solvers into the common volume and compared to the reference output. An error is thrown if they don't match. Since it is possible for the output of the identical tests to differ in floating points due to specifics of the machine and OS, after simple comparison of the output file by file it is also compared numerically using `compare_results.sh`. It strips aways non-numerical data and then compared files field by field with respect to specified relative tolerance.
```
MakisH: Minor language suggestions:
```suggestion
Afterwards, when the simulation finishes, the output is copied from all the solvers into the common volume and is compared to the reference output. An error is thrown if they don't match.

Since it is possible for the output of two identical tests to differ in floating points due to a different machine or OS (as we test different systems and properties), after simple, file by file comparison of the output, the files are also compared numerically using `compare_results.sh`. This script strips away non-numerical data and then compares files field by field with respect to a specified relative tolerance.
```

Is my suggestion `due to a different machine and OS (as we test different systems and properties)` valid?

MakisH: Which is the default tolerance? Where is this defined?

What happens when a test fails in the first comparison but succeeds in the second? Why are we confident that we don't get false successes?

docs/main.md
```diff
@@ -0,0 +1,61 @@
...
+
+In either case after tests finishes, all volumes, networks and containers are removed.
```
MakisH:
```suggestion
In either case, after a test finishes, all the related volumes, networks and containers are removed.
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
+
+# General architecture
```
MakisH: Don't we need a different title here? This is the same as in `main.md`.

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
+
+# General architecture 
+
```
MakisH, line 3: Maybe add a few introductory lines here? Currently the sections feel a bit disconnected in first sight.

docs/main.md
```diff
@@ -0,0 +1,61 @@
+# General architecture 
+
```
MakisH: Maybe add a few lines to describe what this file is and what other files there are under `docs/`?

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+## Triggering the build from the adapter 
+
+In order not to duplicate tests on different adapters, and reuse functionality provided here, Travis API
```
MakisH: Active voice is easier to read in this case:
```suggestion
In order not to duplicate tests on different adapters, and to reuse functionality provided here, we use the Travis API 
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
+
+# General architecture 
+
+## Triggering the build from the adapter 
+
+In order not to duplicate tests on different adapters, and reuse functionality provided here, Travis API 
+is used to trigger selected tests from this repository after commits to the adapters repositories. This is done by
```
MakisH:
```suggestion
to trigger selected tests from this repository after commits to the adapters repositories. This is done by
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+generating custom JSON with specification of the job and sending it to the Travis that triggers the build (For overview of build triggering see [here](https://docs.travis-ci.com/user/triggering-builds/).
```
MakisH: * What is a "JSON with specification of the job"?
* Which is the "Travis (job?) that triggers the build"? Which other Travis ?jobs? do we have? Would be good to document this.

MakisH:
```suggestion
generating custom JSON with specification of the job and sending it to the Travis that triggers the build (For overview of build triggering see [here](https://docs.travis-ci.com/user/triggering-builds/)).
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+Once the tests are finished their return status is queried/sent back (controlled by `--wait` argument) to the Travis job on the adapter repository, making it fail/pass, depending on the return status.
```
MakisH:
```suggestion
Once the tests are finished, their return status is queried/sent back (controlled by `--wait` argument) to the Travis job on the adapter repository, making it fail/pass, depending on the return status.
```

MakisH: "queried" and "sent back" sound like different actions with different subjects here:

* queried: the caller asks the callee (e.g. every some time)
* send back: the callee returns after finishing its task

Which one do we have here?

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+You can learn more from `trigger_systemtests.py`, that handles all job generation and triggering.
```
MakisH:
```suggestion
You can learn more from `trigger_systemtests.py`, which handles all job generation and triggering.
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+
+Since we don't need run full set of tests, when generating JSON only following stages are added:
```
MakisH:
```suggestion
Since we don't need to run a full set of tests, only the following stages are added when generating the JSON job configuration file:
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+- Building the adapter 
+- Running the tests in which it participates
+
+## Running on different branches and pull requests
+
+During the Travis job on the adapter if it is identified, that the job is running on pull request on non-master branch (using
```
MakisH:
```suggestion
During the adapter's Travis job, if it is identified that the job is running on a pull request on a non-master branch (using the 
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+You can learn more from `trigger_systemtests.py`, that handles all job generation and triggering.
+
+Since we don't need run full set of tests, when generating JSON only following stages are added:
+- Building the adapter 
+- Running the tests in which it participates
+
+## Running on different branches and pull requests
+
+During the Travis job on the adapter if it is identified, that the job is running on pull request on non-master branch (using 
+`TRAVIS_BRANCH` and `TRAVIS_PULL_REQUEST` environment variables), the JSON is patched with additional bash script that modifies
```
MakisH: (is actually the JSON file patched with a script?)
```suggestion
`TRAVIS_BRANCH` and `TRAVIS_PULL_REQUEST` environment variables), the JSON file is patched with an additional bash script that modifies
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+the adapter Dockerfiles with additional instruction to switch to the needed branch (obviously the instruction is injected prior
```
MakisH: I would avoid using terms such as "obviously" in documentation, as they are emotionally charged. The term "unsurprisingly" conveys the same message in a more neutral way. However, I would still just skip the qualifier.
```suggestion
the adapter Dockerfiles with additional instruction to switch to the requested branch (prior 
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+- Running the tests in which it participates
+
+## Running on different branches and pull requests
+
+During the Travis job on the adapter if it is identified, that the job is running on pull request on non-master branch (using 
+`TRAVIS_BRANCH` and `TRAVIS_PULL_REQUEST` environment variables), the JSON is patched with additional bash script that modifies
+the adapter Dockerfiles with additional instruction to switch to the needed branch (obviously the instruction is injected prior 
+to actually building an image).
+
+The resultant image is then tagged with the name of the branch/pull request and pushed to the Dockerhub. In the `docker-compose.yml` the image tag
```
MakisH: A link would be useful:
```suggestion
The resultant image is then tagged with the name of the branch/pull request and pushed to [Dockerhub](https://hub.docker.com/u/precice). In `docker-compose.yml`, the image tag 
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Running on different branches and pull requests
+
+During the Travis job on the adapter if it is identified, that the job is running on pull request on non-master branch (using 
+`TRAVIS_BRANCH` and `TRAVIS_PULL_REQUEST` environment variables), the JSON is patched with additional bash script that modifies
+the adapter Dockerfiles with additional instruction to switch to the needed branch (obviously the instruction is injected prior 
+to actually building an image).
+
+The resultant image is then tagged with the name of the branch/pull request and pushed to the Dockerhub. In the `docker-compose.yml` the image tag 
+is controlled using  `{${SOLVER_NAME}_TAG}`( defaults to `latest`). Similarly this environment variables is also specified in the specification of the Travis job
```
MakisH:
```suggestion
is controlled using  `{${SOLVER_NAME}_TAG}`( defaults to `latest`). Similarly, this environment variable is also specified in the specification of the Travis job
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+for every solver participating in the test.
+
+## Adding CI on the adapter
```
MakisH:
```suggestion
## Adding CI for an adapter
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+
+Add specifications of the adapter to the `adapter_info` dictionary in `trigger_systemtests`.
```
MakisH:
```suggestion
Add specifications of the adapter into the `adapter_info` dictionary in `trigger_systemtests.py`.
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+Then the only thing remaining is to add `.travis.yml` to the adapter:
```
MakisH:
```suggestion
Then, the only thing remaining is to add a `.travis.yml` file to the adapter's repository, such as:
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,37 @@
...
+
+    - if: branch = master
+      script: |
+        curl -LO --retry 3 https://raw.githubusercontent.com/precice/systemtests/master/trigger_systemtests.py
+        python3 trigger_systemtests.py --adapter openfoam
+
+We also need to add `TRAVIS_ACCESS_TOKEN` in each adapter repository as a secure variable to authenticate build triggering. It can be found on the [Travis profile page](https://travis-ci.org/account/preferences).
```
MakisH:
```suggestion
We also need to add a `TRAVIS_ACCESS_TOKEN` in each adapter's repository, as a secure variable to authenticate the build triggering. This token can be found on the [Travis profile page](https://travis-ci.org/account/preferences).
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
+# Useful documentation
```
MakisH, line 1: We probably need a document-wide title (heading 1, `#`) and a short introduction.

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+* [Docker](https://docs.docker.com/)
+* [Docker-compose](https://docs.docker.com/compose/)
+* [Main Travis docs](https://docs.travis-ci.com/)
+* [Travis API](https://developer.travis-ci.com/)
+
+
+# Useful patterns and commands
+##  During development of standalone Dockerfiles
+
+If something fails during the build you can determine the command where it failed and comment out the remaining of the Dockerfile. Then rebuild and tag for instance `test_image`. You can then interactively try to follow commented out section of the Dockerfile, by starting the container interactively
```
MakisH:
```suggestion
If something fails during building, you can determine the failing command and comment out the remaining of the Dockerfile. Then, rebuild and tag (e.g. `test_image`). You can then interactively try to follow the commented out section of the Dockerfile, by starting the container interactively:
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+```
+   docker run -it --rm test_image /bin/bash
+```
+Usually in this way or similar it is way easier to debug errors.
```
MakisH:
```suggestion
Usually it is way easier to debug in this way.
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
+# Useful documentation
+
+* [Docker](https://docs.docker.com/)
+* [Docker-compose](https://docs.docker.com/compose/)
+* [Main Travis docs](https://docs.travis-ci.com/)
+* [Travis API](https://developer.travis-ci.com/)
+
+
+# Useful patterns and commands
```
MakisH, line 9: Very good advice!  :+1:  :+1:  :+1:

docs/development.md
MakisH: The second section can actually be the main one and the "useful documentation" could follow as a subsection.

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+```
+   docker run -it --rm test_image /bin/bash
+```
+Usually in this way or similar it is way easier to debug errors.
+
+## During development of the test cases
+
+If something fails during running `docker-compose` due to communication problems or invalid input, the easiest way to debug is:
+
+1. Add sleep timer (e.g. `sleep 1000`) after running failed command
```
MakisH:
```suggestion
1. Add a sleep timer (e.g. `sleep 1000`) after the failing command
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+2. Rerun the failed setup
```
MakisH: It should fail every time
```suggestion
2. Rerun the failing setup
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+3. Attach to the failed container with e.g.
```
MakisH:
```suggestion
3. Attach to the failing container with e.g.
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+```
+   docker-compose run -it calculix-adapter
+```
+4. You can then try to rerun command in `docker-compose.yml` and check the folders `/home/precice/Data/Input`, `/home/precice/Data/Exchange`,
```
MakisH:
```suggestion
4. You can then try to rerun the commands in `docker-compose.yml` and check the folders `/home/precice/Data/Input`, `/home/precice/Data/Exchange`, 
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+2. Rerun the failed setup
+3. Attach to the failed container with e.g.
+```
+   docker-compose run -it calculix-adapter
+```
+4. You can then try to rerun command in `docker-compose.yml` and check the folders `/home/precice/Data/Input`, `/home/precice/Data/Exchange`, 
+`/home/precice/Data/Output`. Make sure the `precice-config.xml` is properly setup, input can be read and folders have proper rights (folders should be
+owned by user `precice`, not the `root` user and not the user of your local machine, unless its `gid` and `uid` matches the one specified in the base Dockerfile)
+
+You should also try to make sure, both runs on local and remote machine works by testing both `local_test.py` and `system_testing.py` with the considered test case.
```
MakisH:
```suggestion
You should also make sure that running both on a local and a remote machine work, testing both `local_test.py` and `system_testing.py` with the considered test case.
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+
+## If something fails on Travis
+
+1. Inspect the build log
+2. Restart the build (maybe it is a connection problem)
+3. Try to replicate it locally with e.g.
+```
+   python3 system_testing.py -s su2-ccx
```
MakisH, line 44: We still need to document somewhere what the difference between `local_test.py` and `system_testing.py` is. Probably in the main README.md file.

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+``` 
+4. If it fails locally, use above-mentioned techniques to debug it. If is does not, bad luck. 
+
+## If there are some network problems
```
MakisH:
```suggestion
## If there are network problems
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+
+It can happen, that you cannot access the network from the inside of docker. You can try to fix it using `--network=host` during build of docker images or by specifying `network: host` in the `docker-compose.yml`.
+
+
+## If CI on adapters does not work
+
+You first need to add `TRAVIS_ACCESS_TOKEN` as environment variable on your system.
+Then you can  run and check generated JSON file without sending a request
```
MakisH:
```suggestion
Then you can  run and check the generated JSON file without sending a request:
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+```
+   python3 trigger_systemtests.py --adapter su2 --test
+``` 
+If it looks fine, you can run it without `--test` to actually send the request to Travis. Then check triggered builds.
```
MakisH:
```suggestion
If it looks fine, you can run it without `--test` to actually send the request to Travis. Then check the triggered builds.
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+If there are no builds visible there, check `requests` section of Travis, maybe Travis could not interpret request correctly.
```
MakisH:
```suggestion
If there are no builds visible there, check the `requests` section of Travis; maybe Travis could not interpret the request correctly.
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+
+## If you want to use Dockefiles for your own development
+
+Some users might want to run coupled simulation without having it installed on your own machine.
```
MakisH:
```suggestion
Some users might want to run coupled simulations without having everything installed on the same system. 
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+```
+   python3 trigger_systemtests.py --adapter su2 --test
+``` 
+If it looks fine, you can run it without `--test` to actually send the request to Travis. Then check triggered builds.
+If there are no builds visible there, check `requests` section of Travis, maybe Travis could not interpret request correctly.
+
+## If you want to use Dockefiles for your own development
+
+Some users might want to run coupled simulation without having it installed on your own machine. 
+For instance mount your own input for each adapter to `/home/precice/Data/Input` and get output from
```
MakisH: Not sure what exactly you mean here. Maybe rephrase it?

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+`/home/precice/Data/Output`.
+
+With the provided setup and a few tweaks it is easy to achieve. On the example of the fenics-adapter:
```
MakisH:
```suggestion
With the provided setup and a few tweaks, this is easy to achieve. Using the example of the fenics-adapter:
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+
+```
+docker build -f Dockerfile.Ubuntu1804.home -t precice-user --build-arg uid=$UID --build-arg gid=$GID . 
+docker build -f Dockerfile.fenics-adapter -t fenics-adapter-user --build-arg from=precice-user
+```
+
+where `GID` and `UID` are you group id (obtained e.g. with `id -a` on Linux).
+
+Then you need to create volume for data exchange and modify `precice-config.xml`, so that data exchange folder
```
MakisH:
```suggestion
Then you need to create a volume for data exchange and modify `precice-config.xml`, so that the data exchange folder 
```

docs/development.md
```diff
@@ -0,0 +1,88 @@
...
+# this regex might differ depending on your precice-config.xml
+sed -i 's|exchange-directory\="."|exchange-directory="/home/precice/Data/Exchange/" network="eth0"|g' \
+    configs/precice-config.xml && cp $tutorial_path/config.yml  configs/config.yml
+```
+
+Then you are free to run your coupled simulation for each participant with:
+```
+docker run --user=precice -it -v $(pwd)/HT/partitioned-heat/fenics-fenics:/home/precice/Data/Input -v exchange:/home/precice/Data/Exchange fenics-adapter-user /bin/bash```
+```
+output can then be copied from the containers.
```
MakisH:
```suggestion
Output can then be copied from the containers.
```

docs/main.md
MakisH: I tried this, it worked out-of-the box! :smile:

docs/adapters-ci.md
MakisH: We also need a note like 

>At the end, remember to document the new test in the README.md and docs/main.md.

MakisH (changes requested):
Thanks for the great work! :+1: :+1: :+1:

Overall, I think that this documentation is very good and very useful! I have added a few comments on problems or questions I had, as well as a few minor language suggestions.

Note that:
* I have not compared the documentation to the code.
* I have not tried to add a new system test

README.md
shkodm: I wanted to keep the main `README.md` minimal, otherwise no one reads it (like the current README...) . But probably overview of the what repository contains is useful. I will also add description of what scripts do.

README.md
shkodm: Hm, interesting, I guess you spotted a bug. I will fix it in a moment.

docs/main.md
shkodm: `README` is a good idea. Though I would not move `main.md` into `README`, rather I would use it to describe different documentation files, as you suggest here.

docs/adapters-ci.md
shkodm: > JSON with specification of the job

This is explained in the provided link to Travis API

> Which is the "Travis (job?) that triggers the build"? Which other Travis ?jobs? do we have? Would be good to document this.

I will elaborate a bit more on this.

docs/adapters-ci.md
shkodm: We have both . It is controlled by `--wait` argument. In one case the job on the adapter Travis query the status (every some time), so it is kind of blocking call if you run with `--wait` flag. Without it it is non blocking, it just triggers the build and returns. If systemtests fails, the return status is sent back, causing to generated "Failed job" on the adapter. I might elaborate more on this in the documentation.

docs/adapters-ci.md
shkodm: Yep...This way prior to building, the script is executed and  dockerfiles get additional commands as wel.

docs/adapters-ci.md
shkodm: > At  the end, remember to document the new test in the README.md and docs/main.md. 

This should be added, but not here probably. Here I describe how to add CI on adapters-repository, given that tests already exists. I should this comment in `main.md` or `development`.

docs/main.md
shkodm: We want the input to different adapters to be in different directories.

docs/main.md
shkodm: Default tolerance is set in the `compare_results.sh` (it can also adjusted via command line args). The general documentation of all numerical comparison is done there, so I don't duplicate it here.

> Why are we confident that we don't get false successes? 

I did some experiments on my machine with changing the results to be above the tolerance and checking if the error is reported and it is. Of course it is still hard to judge what should be the limit of the tolerance, etc, but given the main cause of failed tests are usually building of adapters of some communication problem and rarely numerics,  I would not put much focus here, until we get more failing cases due to numerics.

README.md
shkodm: >> Dockefile

Actually not. It is a typo. Should be `Dockerfile` not `Dockefile`

README.md
MakisH: Keeping it short is good! But, since it's the first document one reads, it should give directions on what to expect and what to read next.

README.md
MakisH: Indeed, tab-filling is not working, so I copied the command. Using the correct file starts smoothly.

The `-d` vs `-f` still holds.

docs/main.md
MakisH: I understoon this later on, but maybe a rephrasing would help, e.g.

> The tutorials are prepared to run in the same environment and several modifications are necessary to run the solvers in different containers and testing conditions: distribute the input files, adjust the simulated time, adjust the solver version, and modify the `precice-config.xml`.

What do you think?

docs/main.md
MakisH: Usually the building part we already test while developing and it is not the/my main reason we rely on automated testing in this case. We really don't want to introduce numerical changes without noticing, so going conservative in this case would help.

I see that in the `compare_results.sh` we have default values:
```
avg_diff_limit="0.01"
max_diff_limit="0.01"
```
as well as values from arguments (which, if I understand correctly, are not used in `system_testing.py`). So,one needs to know that the tests will complain if the results differ more than 1% (but this default threshold is [another discussion](https://github.com/precice/systemtests/issues/90)).

The information that is missing here (and causes some confusion to an adapter developer) is which values the scripts are used with. So, maybe a clarification would help:

> The default value for the tolerance is set in `compare_results.sh` and can be changed, however the script `system_testing.py` (Travis) uses the default values.

docs/main.md
MakisH: That's also good! :+1:

docs/adapters-ci.md
MakisH: Before I understood "we add a script to the JSON file", but I see that you actually mean "we use a script to patch the JSON files". Maybe naming the script would avoid the confusion here (although I agree, that was not a very bright thought).

README.md
```diff
@@ -1,79 +1,33 @@
...
-
-# Status
 [![Build Status](https://travis-ci.org/precice/systemtests.svg?branch=master)](https://travis-ci.org/precice/systemtests)
 
-# Setup and running
+System tests for the distributed multi-physics coupling library library preCICE. 
 
-## Dependencies
+This repository provides Dockerfiles for building preCICE images, solvers, and adapters, as well as docker-compose files and scripts that coordinate everything.
+More information can be found in the [docs](docs/))
```
fsimonis:
```suggestion
More information can be found in the [docs](docs/)
```

README.md
```diff
@@ -1,79 +1,33 @@
...
-
-# travis.yml
-Continuous Integration config.
-Two [build stages](https://docs.travis-ci.com/user/build-stages/):
-- stage 1 (build stage): precice docker image build and pushed to [Docker Hub](https://hub.docker.com/u/precice)
-- stage 2 (testing stage): build su2-ccx, of-of, of-ccx and push outputfiles to repo
-
-# compare_*.py
-Python script to compare reference data with output data.
+    * With ```configparser``` (install for example with ```pip3 install configparser```)
```
fsimonis: Please add a `requirements.txt` and mention `pip3 install --user -r requirements.txt`

README.md
```diff
@@ -1,79 +1,33 @@
...
-Make sure to install 
+## Dependencies
 
-* preCICE (https://github.com/precice/precice/wiki)
-* docker (see https://docs.docker.com/install/linux/docker-ce/ubuntu/)
-    * make sure you can run docker as non-root user (see https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user)
-    * test your docker installation with ```docker run hello-world```
+* [preCICE](https://github.com/precice/precice)
+* [docker](https://docs.docker.com/install/linux/docker-ce/ubuntu/)
+    * Make sure you can run docker as non-root user (see [here](https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user))
```
fsimonis, line 13: Maybe mention to also check weather the default network works as expected by running a ping inside a small container or similar.

README.md
```diff
@@ -1,79 +1,33 @@
...
 
-* preCICE Library path: `-L/home/precice/precice-install/libs` or `pkg-config --libs-only-L libprecice`
-* preCICE Include path: `-I/home/precice/precice-install/include` or `pkg-config --cflags-only-I libprecice`
-* preCICE library flag: `-lprecice` or `pkg-config --libs-only-l libprecice`
 
-# Adding CI to a new adapter
+## Run a system test locally
 
-As described in [#22](https://github.com/precice/systemtests/pull/22),  using `trigger_systemtests.py` it is possible to trigger custom systemtests from the travis job of the another adapter, therefore providing continous integration for the adapter on each commit to a particular branch. If you want to add another adapter, or modify tests that are run for it, you need to modify `nm_repo_map` and `nm_test_map` variables in the `trigger_systemtests.py`, that describe mapping between adapter name and repository and adapter name and set of test cases correspondingly.
+To run OpenFOAM-OpenFOAM system test using Ubuntu 16.04 as the base for building preCICE and adapters:
```
fsimonis, line 29: Please mention `local_test.py -h` for more information.

docs/README.md
```diff
@@ -0,0 +1,5 @@
+This folder provides general documentation of the preCICE system testing architecture.
```
fsimonis: Please add a simple header.

docs/README.md
```diff
@@ -0,0 +1,5 @@
+This folder provides general documentation of the preCICE system testing architecture. 
+
+- General overview of the testing architecture is provided in `architecture.md`. This is a file you should read first.
```
fsimonis:
```suggestion
- General overview of the testing architecture is provided in [`architecture.md`](./architecture.md). This is a file you should read first.
```

docs/README.md
```diff
@@ -0,0 +1,5 @@
...
+- Handling of CI for the individual adapters is described in `adapters-ci.md`
```
fsimonis:
```suggestion
- Handling of CI for the individual adapters is described in [`adapters-ci.md`](./adapters-ci.md)
```

docs/README.md
```diff
@@ -0,0 +1,5 @@
...
+- Some tips and practices for the development are found in `development.md`
```
fsimonis:
```suggestion
- Some tips and practices for the development are found in [`development.md`](./development.md)
```

docs/adapters-ci.md
```diff
@@ -0,0 +1,47 @@
...
+- Running the tests in which it participates
+
+## Running on different branches and pull requests
+
+During the adapter's Travis job, if it is identified that the job is running on a pull request on a non-master branch (using the
+`TRAVIS_BRANCH` and `TRAVIS_PULL_REQUEST` environment variables), the generated JSON filed is patched with additional bash script that modifies
+the adapter Dockerfiles with additional instruction to switch to the needed branch (the instruction should be injected prior
+to actually building an image).
+
+The resultant image is then tagged with the name of the branch/pull request and pushed to [Dockerhub](https://hub.docker.com/u/precice). In the `docker-compose.yml` the image tag
```
fsimonis:
```suggestion
The resulting image is then tagged with the name of the branch/pull request and pushed to [Dockerhub](https://hub.docker.com/u/precice). In the `docker-compose.yml` the image tag
```

docs/architecture.md
```diff
@@ -0,0 +1,64 @@
+# General architecture
+
+When running on Travis tests are executed in the several stages that are outlined below. 
+When it comes to local tests, steps are identical, but the push to the remote dockerhub repository is absent.
+
+## Building preCICE
+
+Several images of preCICE are built, based on different systems (Ubuntu 16.04/18.04, Arch Linux) and different specifications
+of properties of the image ( root/non-root, package, different MPI versions ). This corresponds to the `Dockerfile.$systemname$.${system_spec}`
+Resultant images are then pushed to the [Dockerhub](https://hub.docker.com/u/precice).
```
fsimonis:
```suggestion

Resulting images are then pushed to the [Dockerhub](https://hub.docker.com/u/precice).
```

docs/architecture.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Building solvers
+
+Is done once for every solver, using `Dockerfile.${solver_name}`. Resultant images are then stored on the [Dockerhub](https://hub.docker.com/u/precice).
```
fsimonis:
```suggestion
Is done once for every solver, using `Dockerfile.${solver_name}`. 

Resulting images are then stored on the [Dockerhub](https://hub.docker.com/u/precice).
```

docs/architecture.md
```diff
@@ -0,0 +1,64 @@
...
+docker build --build-arg from=precice/precice-ubuntu1804.home-develop -d Dockerfile.openfoam-adapter -t openfoam-adapter-ubuntu1804.home-develop .
+```
+Note, that due to the fact that each adapter "adapts" differently, the general procedure is slightly different and is outlined below:
+
+- **SU2 adapter**
+  Since the SU2 adapter modifies the SU2 source code, it makes no sense to copy a clean SU2 build, because it will need to be rebuilt. Instead we directly build adapted
+  SU2 source.
+
+- **deal.II adapter**
+  Since deal.II is a library and the adapter just needs to link with it, here we actually perform the import of deal.II image and only build
```
fsimonis:
```suggestion
  Since deal.II is a library and the adapter just needs to link to it. Thus, we perform the import of deal.II image and only build
```

fsimonis (changes requested):
There are some cosmetic and grammatical issues.

README.md
shkodm: Actually, it is good that you noticed. We don't actually use this dependency. I will remove it.

docs/development.md
shkodm: Yes. Explanation that I wrote above should be in `docs/development.md` now. I also added some motivation behind the solution in `docs/architecture.md`.

shkodm:
As discussed with @IshaanDesai on gitter, he will report on the problems with documentation from the user perspective in the future. Documentation will also slightly be updated when new changes are introduced. I merge this for now.

PR #88, opened 2019-08-29
Organize scripts, base images  and tests into folders

shkodm:
Over time file structure became a mess - Dockerfiles for preCICE, Dockerfiles for adapters, Dockerfiles for solvers, scripts and tests all in root directory... This fixes it by introducing following directory structure:

**base_precice** - dockerfiles for preCICE
**base_solvers** - dockerfiles for building standalone solver (no preCICE)
**base_adapters** - dockerfiles for building adapted solvers
**tests** - no explanation needed

I keep scripts in the root directory for convenience.


fsimonis:
Let's just call them:

* precice
* adapters
* solvers

PR #95, opened 2019-09-12
Run test_install

BenjaminRodenberg:
If we install preCICE using `make install`, we should also test the installation like described [here](https://github.com/precice/precice/wiki/Tests#testing-the-installation).


BenjaminRodenberg:
@fsimonis do you think this PR is still relevant? If yes, I would suggest to merge it, if all tests are successful. I will trigger the tests as soon as #123 is merged (otherwise they will not work).

fsimonis:
I think that this test is unnecessary as we control the complete environment in the system tests.
Basically every system tests is an installation test.

PR #96, opened 2019-09-18
Fix user for binding tests

BenjaminRodenberg:
We should use `USER precice` for the bindings tests, as well.

This PR should only be merged after #123 has been merged.


BenjaminRodenberg:
I observed that `sudo apt install python3-pip` does not work for Ubuntu 18.04, since the version is too old (see https://github.com/precice/precice/wiki/Non%E2%80%93standard-APIs#python-bindings-version-of-pip3-is-too-old). We have to install `pip3` using `wget ... && python3 get_pip.py` (currently installs: version 19.2.3) instead.

I believe that at an earlier point in time `pip3` provided via `apt` was sufficient (pip3 version 9.0.1 on Ubuntu1804). This was also the original reason for providing `Test_bindings` and `Test_bindings.Ubuntu1804`. We obviously have a regression here...

I do not know, when we actually introduced changes that broke compatibility with pip 9.0.1. I am also not 100% sure whether we should try to make the python bindings compatible with pip 9.0.1, again, or whether the current state is fine (for sure running `apt install pip3` is simpler than the current approach...). 

 :question: **Any opinions on this question** :question: 

I am currently investigating the situation if I use Ubuntu 19.04. *Edit:* See #97

BenjaminRodenberg:
@shkodm Something else seems not to work with the runner scripts: https://travis-ci.org/precice/systemtests/jobs/586477760#L445-L452

BenjaminRodenberg:
On Ubuntu 1904, we can actually use the pip provided via apt. See #97

shkodm (approved):

shkodm:
> @shkodm Something else seems not to work with the runner scripts: https://travis-ci.org/precice/systemtests/jobs/586477760#L445-L452

Nice catch. I will fix that.

fsimonis (approved):

Issue #102, opened 2019-10-13
Fix test Ubuntu1804.sudo.mpich


BenjaminRodenberg:
Currently, this test is failing: See https://travis-ci.org/precice/systemtests/jobs/597143039

The last successful run of this test that I could find is https://travis-ci.org/precice/systemtests/jobs/583929546.

If this test fails, the following stages do not even get executed. This is bad, since this means that the adapters are not built and the actual system tests (important tests) are not even being run. Note that these important tests do not rely on Ubuntu1804.sudo.mpich. This means we could actually execute these tests, even if Ubuntu1804.sudo.mpich is failing.

There are two ToDos in this issue:

- [ ] fix the test. Why is it failing? If necessary, open an issue in https://github.com/precice/precice
- [ ] remove stage-dependency of important tests on previous stages, if there is no actual dependency

MakisH:
We first need to understand what the actual error is. Currently, the `test_base` fails, with the following log:

```
1/3 Test: [secure].Base
Command: "/usr/bin/mpiexec.mpich" "-n" "4" "/home/[secure]/[secure]-build/test[secure]" "--run_test=!@MPI_Ports:!MappingTests/PetRadialBasisFunctionMapping/Parallel/*"
Directory: /home/[secure]/[secure]-build/TestOutput/Base
"[secure].Base" start time: Oct 26 10:29 UTC

Output:
----------------------------------------------------------

Running Running 268 test cases...
Running 263 test cases...
Running 263 test cases...
263 test cases...

(0) 10:30:03 [mapping::NearestProjectionMapping]:126 in computeMapping: WARNING: 3D Mesh "InMesh" does not contain triangles. Nearest projection mapping will map to primitives of lower dimension.(1) 10:30:03 [mapping::NearestProjectionMapping]:126 in computeMapping: WARNING: 3D Mesh "InMesh" does not contain triangles. Nearest projection mapping will map to primitives of lower dimension.

(0) 10:30:03 [mapping::NearestProjectionMapping]:126 in computeMapping: WARNING: 3D Mesh "InMesh" does not contain triangles. Nearest projection mapping will map to primitives of lower dimension.
(3) 10:30:03 [mapping::NearestProjectionMapping]:126 in computeMapping: WARNING: 3D Mesh "InMesh" does not contain triangles. Nearest projection mapping will map to primitives of lower dimension.
[...]

PR #108, opened 2019-10-16
Add `allow_failures` section to travis.yml

Eder-K:
Adds a new section of jobs in the travis.yml file that do not terminate the build after the current stage finishes. Also adds the currently failing Ubuntu1804.sudo.mpich job to this section to allow further stages to run (#102).

For a job to be allowed to fail, it must be specified in _both_ the `include` and `allow_failures` sections. It done so, it will not cause subseqent stages to be cancelled.

**Note**: If every job except for `allow_failures` ones succeeds, the full build will still be marked as successful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ services:
 
 jobs:
   allow_failures:
-  - name: "Ubuntu 18.04.sudo.mpich"
+  - name: "Ubuntu 18.04.sudo.mpich [Job failure permitted]"
     if: type = cron
     script:
     - docker build -f precice/Dockerfile.Ubuntu1804.sudo.mpich -t precice/precice-ubuntu1804.sudo.mpich-develop .
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,17 @@ python:
   - "3.5"
 services:
   - docker
+
 jobs:
+  allow_failures:
+  - name: "Ubuntu 18.04.sudo.mpich"
+    if: type = cron
+    script:
+    - docker build -f precice/Dockerfile.Ubuntu1804.sudo.mpich -t precice/precice-ubuntu1804.sudo.mpich-develop .
+    after_success:
+    - echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin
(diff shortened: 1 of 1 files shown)


MakisH:
So, we just need to make sure that we look at the detailed view once in a while? Then a good question would be what should be allowed to fail and why.

Eder-K:
I think that in general, we should not designate builds as allowed failures, unless a failing build is temporarily hindering the regular checks of tests in higher up stages, as is currently the case for sudo.mpich #102 .

---
When I was thinking of ways on how to fix failing lower stage build blocking higher stage tests I found that any solution requires a decision around the question

'_How crucial are stages to the build process?_`

Stages by definition are meant to block later stages if not all stages before that were executed successfully. However, by grouping all precice builds into a single stage we implicitly demand that any failure of precice on any of the defined systems should halt adapter builds and actual tests.
So if we want to keep this order of stages and run all stages, we have to mark currently failing builds in early stages as allowed failures.

Another way to do this would be to remove stages entirely. Then all jobs are executed independently and the failure of a job like sudo.mpich will not interfere with more crucial tests.
But then failures of critical builds like Ubuntu1804.home/Ubuntu1604.home will not halt the tests (almost every test uses one of these).

---
Unfortunately, I found no functionality in travis to establish more complex dependencies between jobs, so the options seem to be 'use stages or don't use stages'.
A compromise of sorts would be to move the less important builds (mainly the cron specific builds) to a fourth stage after the actual tests. This preserves the general stage structure of `precice build` - `adapter build` - `test` and the full travis build will still fail if any job fails.

That way we would not need to designate tests as allowed to fail at all.

Eder-K:
After discussing in person with @MakisH , we have chosen to add the 'allowed to fail' section and move tests there to not disrupt the overall stage structure.

To make it more clear that certain tests are allowed to fail (and will not affect the end status of the full build), the tag `[Job failure permitted]` has been added in the title of the job to highlight this.

Other jobs that are temporarily marked as allowed failures in the future should bear a similar tag to notify users that the exit status of this particular job might not reflect the status of the full travis build.

PR #111, opened 2019-10-18
Modifying fenics - fenics test case

IshaanDesai:
This PR modifies the fenics-fenics test case to incorporate the design changes made in https://github.com/precice/tutorials/pull/46 and also test the vector implementation in https://github.com/precice/fenics-adapter/pull/46 and https://github.com/precice/fenics-adapter/pull/47. 
While testing all these features manually it was observed that the adapter `referenceOutput` is changed due to better performance. Reference output of this test case has been correspondingly modified. 
**NOTE**: This PR only modifies the test case for the simple interface (interface is parallel to y-axis). https://github.com/precice/tutorials/pull/46 also incorporates complex domain analysis capabilities for which another test needs to be devised


BenjaminRodenberg (approved):
Looks good. If the reference results get better, then this is a good thing :+1:

Eder-K (approved):
Agreeing with @BenjaminRueth , well done.

PR #116, opened 2019-10-31
Perform `after_success` actions in `deploy` section instead

### LINKED ISSUE #103: Test should fail, if pushing fails
In [this test](https://travis-ci.org/precice/systemtests/jobs/597143037#L2681-L2683) pushing to dockerhub fails, but the test succeeds. I think that this is not what we want. Is there a way to throw an error or similar, if pushing to docker hub fails, such that Travis can mark the test as failing?

Eder-K:
Closes #103 .

---
For our systemtests builds, we intend that after each job finishes, some form of result will be pushed to an external hub (DockerHub or `precice_st_output`) based on the success of the job. This is currently done by specifying `after_success` and `after_failure` keys in the `.travis.yml` configuration file with the relevant pushing commands.

However, TravisCI [by design](https://docs.travis-ci.com/user/job-lifecycle/#breaking-the-build) does not factor in the success of these post-job commands when expressing the exit status of the build. In the case of a failure in the `after_success` commands, the build will still be marked as success (jobs executing the `after_failure` sections will rightfully get marked as failures).

---
To circumvent this issue, we can instead designate `after_success` commands as `deploy` commands. These sections are only executed if the job succeeds and will cause the job to be marked as failed if the deploy fails.
This PR replaces all current `after_success` sections with deployments and adds documentation on how to add tests and handle `after_failure` and `deploy` sections.

Note: If a job fails during its main commands (e.g. while building an adapter) the build will get marked as '**failed**'. If a job fails during its deployment phase, it gets marked as '**errored**' instead.

### How to use `deploy`

For adding jobs in the future it should be noted that compared to `after_success` sections, deploys require a few additional arguments to function properly. Here is an code snippet from one of the jobs:
```bash
    - stage: Building preCICE
      name: "Arch Linux"
      if: fork = false
      script:
        - docker build -f precice/Dockerfile.Arch -t precice/precice-arch-develop .
      deploy:
        skip_cleanup: true
        provider: script
        on:
          all_branches: true
        script: >-
          echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin &&
          docker push precice/precice-arch-develop:latest
```
In addition to the actual deploy script, the following arguments must be provided inside of the `deploy` key:
- `skip_cleanup: true` - this prevents TravisCI from stashing all changes made during the build, i.e. cleaning up our actual results.
- `provider: script` - specifies that the deployment happens through our manually provided script.
- `on: all_branches: true` - the deployment can be issued on any branch. This allows us to execute commands from the same branch that triggered a travis build, for any branch. Otherwise the deployment scripts would be run exclusively on `master`, independent of which branch executed the main script.
- `script: >-   [actual push script]` - The `>-` is *critical* here. For one, **deploy scripts can only consist of a single line**. This means that we have to chain commands together (note the `&&` at line end) and use `>` to extend them into multiple lines. The addition of the hyphen `-` prevents the last command to be trailed by a newline character `/n`. **If this hyphen is omitted, the `/n` character in the command might get parsed as only `n`, which appends a character to the final command and makes it invalid!**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,223 +8,363 @@ services:
 
 jobs:
   allow_failures:
-  - name: "Ubuntu 18.04.sudo.mpich [Job failure permitted]"
-    if: type = cron
-    script:
-    - docker build -f precice/Dockerfile.Ubuntu1804.sudo.mpich -t precice/precice-ubuntu1804.sudo.mpich-develop .
-    after_success:
-    - echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin
-    - docker push precice/precice-ubuntu1804.sudo.mpich-develop:latest
+    - name: "Ubuntu 18.04.sudo.mpich [Job failure permitted]"
+      if: type = cron
+      script:
+        - docker build -f precice/Dockerfile.Ubuntu1804.sudo.mpich -t precice/precice-ubuntu1804.sudo.mpich-develop .
+      deploy:
+        skip_cleanup: true
+        provider: script
+        on:
+          all_branches: true
+        script: >-
+          echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin &&
+          docker push precice/precice-ubuntu1804.sudo.mpich-develop:latest
 
   include:
-  - stage: Building preCICE
-    if: fork = false
-    name: "Arch Linux"
-    script:
-    - docker build -f precice/Dockerfile.Arch -t precice/precice-arch-develop .
-    after_success:
-    - echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin
-    - docker push precice/precice-arch-develop:latest
+    - stage: Building preCICE
+      name: "Arch Linux"
+      if: fork = false
+      script:
+        - docker build -f precice/Dockerfile.Arch -t precice/precice-arch-develop .
+      deploy:
+        skip_cleanup: true
+        provider: script
+        on:
+          all_branches: true
+        script: >-
+          echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin &&
+          docker push precice/precice-arch-develop:latest
 
-  - stage: Building preCICE
-    if: fork = false
-    name: "Ubuntu 16.04 home"
-    script:
-    - docker build -f precice/Dockerfile.Ubuntu1604.home -t precice/precice-ubuntu1604.home-develop .
-    after_success:
-    - echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin
-    - docker push precice/precice-ubuntu1604.home-develop:latest
+    - stage: Building preCICE
+      name: "Ubuntu 16.04 home"
+      if: fork = false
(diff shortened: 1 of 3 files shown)


MakisH (approved):

.travis.yml
```diff
@@ -8,223 +8,363 @@ services:
...
-    script:
-    - docker build -f precice/Dockerfile.Ubuntu1804.sudo.mpich -t precice/precice-ubuntu1804.sudo.mpich-develop .
-    after_success:
-    - echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin
-    - docker push precice/precice-ubuntu1804.sudo.mpich-develop:latest
+    - name: "Ubuntu 18.04.sudo.mpich [Job failure permitted]"
+      if: type = cron
+      script:
+        - docker build -f precice/Dockerfile.Ubuntu1804.sudo.mpich -t precice/precice-ubuntu1804.sudo.mpich-develop .
+      deploy:
```
Eder-K: This deploy key replaces the after success section.

.travis.yml
```diff
@@ -8,223 +8,363 @@ services:
...
+        skip_cleanup: true
+        provider: script
+        on:
+          all_branches: true
+        script: >-
+          echo "$DOCKER_PASSWORD" | docker login -u precice --password-stdin &&
+          docker push precice/precice-ubuntu1804.sudo.mpich-develop:latest
```
Eder-K, line 22: This deploy block would replace the former `after_success` section.

The remaining changes in this file are the same as above applied to all other jobs.

.travis.yml
```diff
@@ -8,223 +8,363 @@ services:
...
```
Eder-K: The remaining changes in this file are the same as above applied to all jobs.

docs/adding_tests.md
```diff
@@ -0,0 +1,82 @@
+# How to add tests
```
Eder-K, line 1: I wanted to give a full explanation on how to introduce a new test with the new deployment method. However, I felt like this does not quite fit the topics covered in the other documentation files. Therefore I decided on moving this explanation to a new file alltogether.

docs/architecture.md
```diff
@@ -1,7 +1,7 @@
 # General architecture
 
-When running on Travis tests are executed in the several stages that are outlined below. 
-When it comes to local tests, steps are identical, but the push to the remote dockerhub repository is absent.
+When running on Travis tests are executed as specified in the file `.travis.yml` located in the root of the systemtests repository. The several stages of testing are outlined below.
+If running a local tests (compared to the standard build on the Travis servers) the steps are identical, but the push to the remote DockerHub repository is absent.
```
Eder-K, line 4: Referring to the `travis.yml` file here to make it clearer how tests are orchestrated.

BenjaminRodenberg (approved):
Looks good. I checked it from the perspective of the person who opened issue #103 and this solves the issue. Thanks for also providing documentation on this feature!

PR #124, opened 2019-11-27
Use branch develop as default for all tests

BenjaminRodenberg:
We want to test against the develop version of the branches, not against the master version. Therefore, we explicitly add the option `--branch` for all tests where we `git clone` preCICE/an adapter/a tutorial.

### Why do we do this?

Currently, we cannot really test breaking changes without introducing the breaking changes to the whole toolchain. We agreed on allowing breaking changes on the `develop` branches of all `precice/` repositories (thus we can ensure compatibility). 

In order to avoid that our users have to deal with these breaking changes before the official release, we make sure to only update `master`for all `precice/` repositories via a PR from `develop` as soon as the new preCICE version (including the breaking changes) is officially released.


fsimonis (approved):
This is not a breaking change, so lets merge it.

Issue #121, opened 2019-11-25
Add user friendly interface to specify branch for testing


BenjaminRodenberg:
Depending on the test you want to run, different branches of the components needed for the test might be needed. One possible use-case is that one is developing a new feature on a branch of the FEniCS adapter.

For preCICE there already exists an option `--branch` to specify the branch being used in a test. However, if you want to use a different branch than the default one, there is no user friendly way to do this. 

### Example

We want to use the branch `feature` of the FEniCS adapter instead of the default branch `master` for the test `fe-fe`. We only want to run the test locally on our machine. 

1. We have to modify the `Dockerfile` correspondingly.
2. Make sure that the modified `Dockerfile` is used to build the image `precice/fenics-adapter-ubuntu1804.home-develop`.
3. The modified image will then be used in the test (see [here](https://github.com/precice/systemtests/blob/ec4ef9d4aedd0087dfb3a8ed98fdf7a1267c7751/tests/TestCompose_fe-fe.Ubuntu1804.home/docker-compose.yml#L19)), if we run the test locally.

BenjaminRodenberg:
One possible solution: Don't provide only a branch via `--branch`, but also allow to provide a `json` file that describes which components of the test have to use which branch:

```
{'precice':'master', 'fenics-adapter':'feature', 'tutorials':'use-feature'}
```

**Warning:** We should really provide a file and not a `json`-formatted string via command line. Passing such a string via command line will drive everybody crazy w.r.t properly escaping special characters.

PR #126, opened 2019-11-27
Use branch in build_adapters

fsimonis:
This PR changes the systemtests, so that they use the same branch in both:
* the preCICE base image, and
* the adapter image.

This is a first step aimed at making the execution more consistent.
The next step its to implement #121.


BenjaminRodenberg:
Is this PR still relevant?

PR #127, opened 2019-11-28
Add test for elastictube1d

### LINKED ISSUE #16: Add system test for precice/elastictube1d example
The [elastictube1d](https://github.com/precice/elastictube1d) example is simple and easy to run. It would be a nice system test. Additionaly: it would test the python bindings

- [ ] check that the example can be build and run
    - [ ] in C
    - [ ] in python
- [ ] create reference results and compare test result
    - [ ] in C
    - [ ] in python

BenjaminRodenberg:
Closes #16 

We should make sure to test both variants of the tube (C and Python)


BenjaminRodenberg:
The code base for this PR is outdated. Refer to #128.

PR #123, opened 2019-11-26
remove old python bindings

BenjaminRodenberg:
Ensures compatibility to https://github.com/precice/precice/pull/576

These PRs should be merged together:

- [x] https://github.com/precice/fenics-adapter/pull/55
- [x] https://github.com/precice/precice/pull/576
- [x] https://github.com/precice/tutorials/pull/58
- [x] https://github.com/precice/systemtests/pull/123

PR #131, opened 2019-11-28
Add python script for building images and pushing them

### LINKED ISSUE #77: Confusion with lower/upper case branches and tags in docker
Docker only allows lower case tags. Since we are using tags for identifying the branches or baseimages, we have to convert mixed case names to lower case names. This leads to a easy-to-break composition of tags:

https://github.com/precice/systemtests/blob/bf39313461b899b49bc47b5e4ad443259d5bfe87/.travis.yml#L19-L26

At some points, this conversion is already handled properly by our python scripts:

https://github.com/precice/systemtests/blob/bf39313461b899b49bc47b5e4ad443259d5bfe87/system_testing.py#L107

However, I think the current case is pretty confusing. See example above, where one has to know how the branch being used (`develop`) and the name of the Dockerfile (`-f Dockerfile.Ubuntu1604.home`) has to be translated to a tag (`-t $DOCKER_USERNAME/precice-ubuntu1604.home-develop`).

**My suggestion:** Don't use the command `docker build` for building the base images at all, but use something similar to [`system_testing.build`](https://github.com/precice/systemtests/blob/bf39313461b899b49bc47b5e4ad443259d5bfe87/system_testing.py#L19-L31) such that a python script can handle tagging and take care of cases. In the end this will allow us to use something like

```
python build_base.py -f Dockerfile.Ubuntu1604.home -b develop
```

instead of 

```
docker build -f Dockerfile.Ubuntu1604.home -t $DOCKER_USERNAME/precice-ubuntu1604.home-develop .
```

BenjaminRodenberg:
closes #77 

This PR should only be merged after #123 has been merged.


BenjaminRodenberg:
We should not only build the precice images using a python script, but also the other components, that are being reused by other parts of the tests. This means, we should also build the adapters and solvers via a python script:

https://github.com/precice/systemtests/blob/d9152323653037bcb5b57c354d96eb53215488b3/.travis.yml#L151

this line should be changed to

```
python build_adapter.py --dockerfile adapters/Dockerfile.su2-adapter
```

Note that even though `build_base.py` and `build_adapter.py` will have very similar functionality, we should still use different scripts. Currently our tagging system is not 100% consistent, when it comes to adapters and precice base images and therefore, we cannot easily create the correct tag only from the path of the dockerfile:

* for precice images dockerfile is stored at `precice/...`, tag is `precice/precice-...`
* for adapter images dockerfile is stored at `adapters/Dockerfile.su2-adapter` tag is `precice/su2-adapter-...`

The easiest approach to circumvent this is provide independent scripts for each case. Later (different PR), we should rename and restructure our dockerfiles in order to have a consistent system.

BenjaminRodenberg:
Before merging, we should make sure that [Ubuntu 16.04 home PETSc](https://travis-ci.org/precice/systemtests/jobs/623803789?utm_medium=notification&utm_source=github_status) is working again.

build_adapter.py
```diff
@@ -0,0 +1,54 @@
...
+    if args.operating_system:  # first feature is mandatory always describes the os
+        precice_base_features["os"] = args.operating_system
+    if args.precice_installation:  # second feature is optional if it exists it describes the preCICE installation
+        precice_base_features["installation"] = args.precice_installation
+    if args.petsc == "yes":
+        precice_base_features["petsc"] = "yes"
+
+    precice_base_tag = system_testing.compose_tag(args.docker_username, "precice", precice_base_features, args.branch)
+
+    print("building {} image with the following features:{}".format(adapter_name, features))
```
Eder-K:
```suggestion
    print("Building {} image with the following features: {}".format(adapter_name, features))
```

build_precice.py
```diff
@@ -0,0 +1,45 @@
...
+            features["installation"] = feature
+        if i >= 3:  # third and following features are optional
+            assert(feature in ["mpich"])  # we expect one of these optional features
+            if feature is "mpich":
+                features["mpich"] = "yes"
+    assert(i > 0)  # at least one feature (the os) should have been provided
+    if args.petsc == "yes":
+        features["petsc"] = "yes"
+
+    print("building preCICE image with the following features:{}".format(features))
```
Eder-K:
```suggestion
    print("Building preCICE image with the following features: {}".format(features))
```

build_adapter.py
```diff
@@ -0,0 +1,54 @@
+import argparse, docker
+import system_testing
+import os
+                           
+if __name__ == "__main__":
+    # Parsing flags
+    parser = argparse.ArgumentParser(description='Build local.')
```
Eder-K:
```suggestion
    parser = argparse.ArgumentParser(description='Build local docker image of an adapter.')
```

build_precice.py
```diff
@@ -0,0 +1,45 @@
+import argparse, docker
+import system_testing
+import os
+                           
+if __name__ == "__main__":
+    # Parsing flags
+    parser = argparse.ArgumentParser(description='Build local.')
```
Eder-K:
```suggestion
    parser = argparse.ArgumentParser(description='Build local docker image of preCICE.')
```

push_adapter.py
```diff
@@ -0,0 +1,36 @@
+import argparse, docker
+import system_testing
+import os
+                           
+if __name__ == "__main__":
+    # Parsing flags
+    parser = argparse.ArgumentParser(description='Build local.')
```
Eder-K:
```suggestion
    parser = argparse.ArgumentParser(description='Push local adapter image to Docker Hub.')
```

push_precice.py
```diff
@@ -0,0 +1,39 @@
+import argparse, docker
+import system_testing
+import os
+                           
+if __name__ == "__main__":
+    # Parsing flags
+    parser = argparse.ArgumentParser(description='Build local.')
```
Eder-K:
```suggestion
    parser = argparse.ArgumentParser(description='Build local preCICE image to Docker Hub.')
```

Eder-K (approved):
Just some small clarifications for the descriptions. The changes look great though. :+1:

BenjaminRodenberg:
> Just some small clarifications for the descriptions. The changes look great though. +1

Thanks for the review! I applied your suggestions. As soon as the tests have run through successfully, feel free to squash & merge.

system_testing.py
```diff
@@ -168,6 +168,41 @@ def build_run_compare(test, tag, branch, local_precice, force_rebuild, rm_all):
...
+        I.e. the image contains a build of precice.
+
+    FEATURES are ubuntu1604 and home.
+        I.e. ubuntu1604 is used as operating system and preCICE is installed in a user directory
+
+    BRANCH is develop.
+        I.e. image is based on the source code provided at precice:develop, https://github.com/precice/precice/tree/develop.
+    """
+    features_list = []
+    features_list.append(features["os"]) if "os" in features else None
```
fsimonis: You can you `features.get("os")` which returns `None` if the key is missing.

fsimonis (approved):

system_testing.py
BenjaminRodenberg: Thanks for the hint! I added the changes in https://github.com/precice/systemtests/pull/131/commits/ecc0310f5901d0ecbfcfc18993c5e88e187e2c53. Unfortunately for `"petsc"` and `"mpich"` I could not entirely get rid of the `else None` parts.

BenjaminRodenberg: There is a slight problem with using `features.get(key)` instead of `features_list.append(...) if key in features else None`: The first approach appends `None` to the list, if `key` is not found, the second approach does nothing. Therefore, we have to filter away all the `None` elements as a postprocessing step, like introduced in https://github.com/precice/systemtests/pull/131/commits/56ba8af122ad5e00002cfb4e51ffb0d66d98c671.

BenjaminRodenberg: This is not ideal, but I still prefer using `features.get(key)` over the other option, since it is clearer from my point of view.

Eder-K:
Since TravisCI does not execute deployment phases for pushes to PR branches, we have to manually trigger buillds on the website if we want to test the deploy scripts.
I just triggered such a build [here](https://travis-ci.org/precice/systemtests/builds/650745358?utm_medium=notification&utm_source=github_status). If this succeeds, we should be able to merge this without worry.

Update: the nutils-of test did fail. This is not relevant for this PR, however, and all deployments succeeded.

PR #132, opened 2019-12-01
Use Ubuntu18.04 for Nutils-OF test

BenjaminRodenberg:
#123 has to be merged first.

Fixes `nutils-of` test by using preCICE baseimage based on Ubuntu 18.04.

**Background:**

Previously a baseimage based on Ubuntu 16.04 has been used. However, this produces the [following error](https://travis-ci.org/precice/systemtests/jobs/603688857#L757-L786), when trying to `import nutils`. This error can be avoided by using python3 version 3.5.3 or newer (see https://github.com/python/typing/issues/266). Ubuntu 16.04 only provides version 3.5.2. Therefore, we use Ubuntu 18.04 for this test.


uekerman:
According to Gertjan, it is a Python bug, which was fixed in 3.5.3. Unfortunately, Ubuntu 16.04 seems to no longer update to this package. Directly changing to 18.04 is the best and easiest solution for us, I agree.

BenjaminRodenberg:
> According to Gertjan, it is a Python bug, which was fixed in 3.5.3. Unfortunately, Ubuntu 16.04 seems to no longer update to this package. Directly changing to 18.04 is the best and easiest solution for us, I agree.

So this means the approach how we fix the bug is clear: run systemtests that use nutils on Ubuntu 18.04. This also needs a matching version of the OpenFOAM adapter. 

I already implemented this. However there still seems to be a problem which leads to a failure of the test. If anybody knows what's going wrong, feel free to comment.

BenjaminRodenberg:
*Note:* I think [this](https://travis-ci.org/precice/systemtests/jobs/590477900?utm_medium=notification&utm_source=github_status) is the youngest successful test `nutils-of`.

BenjaminRueth:
*Note:* I think [this](https://travis-ci.org/precice/systemtests/builds/590477881?utm_source=github_status&utm_medium=notification) is the youngest successful test `nutils-of`.

BenjaminRodenberg:
The test is working again. @fsimonis @MakisH Short summary of the necessary steps:

* Use Ubuntu 18.04 and OpenFOAM 5 for running the tests (previously Ubuntu 16.04 and OpenFOAM4). We have to use Ubuntu 18.04 due to a bug in the python version that is shipped with Ubuntu 16.04, which led to problems in nutils. We use OpenFOAM 5, since OpenFOAM 4 is not available for Ubuntu 18.04.
* Update reference results for the test due to minor numeric changes.
* Updated version of the test requires us to provide also a dockerimage for the OpenFOAM adapter with Ubuntu 18.04 and for OpenFOAM 5 with Ubuntu 18.04

tests/TestCompose_nutils-of.Ubuntu1804/referenceOutput/OpenFOAM/0.2/T
```diff
@@ -1,9 +1,9 @@
 /*--------------------------------*- C++ -*----------------------------------*\
-| =========                 |                                                 |
-| \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox           |
-|  \\    /   O peration     | Version:  4.1                                   |
-|   \\  /    A nd           | Web:      www.OpenFOAM.org                      |
-|    \\/     M anipulation  |                                                 |
+  =========                 |
+  \\      /  F ield         | OpenFOAM: The Open Source CFD Toolbox
+   \\    /   O peration     | Website:  https://openfoam.org
+    \\  /    A nd           | Version:  6
```
MakisH: This test uses the APT package `openfoam5`, right? Why do the headers show 6?

BenjaminRodenberg: Good point. The machine where I generated the reference results uses OF6. I will rerun the examples with OF5.

MakisH: Apart from the different headers, I also see minor differences in the results, which should not be a problem, since we are now comparing the results with some tolerance. But if you have time, it would be nice to rerun with the same OpenFOAM version. :+1:

MakisH (commented):
Why do we need to replace the configuration files in the first place?

tests/TestCompose_nutils-of.Ubuntu1804/referenceOutput/OpenFOAM/0.2/T
BenjaminRodenberg: I rerun the experiments and added new reference results in https://github.com/precice/systemtests/pull/132/commits/93e79b3ae57e1773e06561b2de707c7305087314. I hope this resolves this point.

BenjaminRodenberg: Regarding the minor differences: Since we do not compare with an analytical solution or experimental results as far as I understand minor changes are ok, as long as the results still "look" good. I checked the results in paraview and could not find any problems.

BenjaminRodenberg:
> Why do we need to replace the configuration files in the first place?

This is what comes from the tutorials: https://github.com/precice/tutorials/blob/develop/CHT/flow-over-plate/buoyantPimpleFoam-nutils/OpenFOAM/precice-adapter-config.yml

BenjaminRodenberg:
@fsimonis note https://github.com/precice/systemtests/issues/134. I think we can deal with this issue in a different PR. I am confident that the error thrown by `awk` does not affect the outcome of the test. I will also check this by provoking a failing test (i.e. putting some wrong data into the reference results) to make absolutely sure.

Eder-K:
Right now the TravisCI build fails due to the Travis results not matching the reference results. I uploaded the Travis results of this Nutils-OF test version on a [branch of the st_output repository](https://github.com/precice/precice_st_output/tree/EderK-push_files/testOutput/Output), if anyone wants to look into them. 
To me they look fine (in that the numerical differences don't change the solution behavior), so we should be able to substitute them for the current reference results. But please feel free to check for yourself to be sure.

BenjaminRodenberg:
> Right now the TravisCI build fails due to the Travis results not matching the reference results. I uploaded the Travis results of this Nutils-OF test version on a [branch of the st_output repository](https://github.com/precice/precice_st_output/tree/EderK-push_files/testOutput/Output), if anyone wants to look into them.
> To me they look fine (in that the numerical differences don't change the solution behavior), so we should be able to substitute them for the current reference results. But please feel free to check for yourself to be sure.

Thanks for taking care of this! I will check the results.

BenjaminRodenberg:
With the new reference results, the test runs through successfully (#142). From my perspective everything is now ready to merge.

PR #142, opened 2019-12-19
Use reference results from precice_st_output

BenjaminRodenberg:
https://github.com/precice/precice_st_output/tree/EderK-push_files/testOutput/Output

provides reference results for #132


BenjaminRodenberg:
New reference results seem to work. Some other tests are making problems, but this is not an issue of this PR.

PR #143, opened 2019-12-19
Allow user to specify branch of systemtests being triggered by trigger_systemtests.py

### LINKED ISSUE #141: Make sure that trigger_systemtest.py uses the correct branches
`trigger_systemtest.py` currently always uses `systemtests:master`. This leads to failure of some tests (for example [fenicsadapter](https://travis-ci.org/precice/fenics-adapter/builds/618292395?utm_source=github_status&utm_medium=notification) triggers [this test](https://travis-ci.org/precice/systemtests/builds/625559439)).

We need a way to specify the branch that `trigger_systemtest.py` should use (`--branch` argument that defaults to `master`) and automatically choose the correct branch in the repository that triggers the test.

BenjaminRodenberg:
closes #141

### DIFF
--- a/trigger_systemtests.py
+++ b/trigger_systemtests.py
@@ -238,7 +238,7 @@ def trigger_travis_and_wait_and_respond(job_body, user, repo):
         exit(1)
 
 
-def generate_failure_callback():
+def generate_failure_callback(branch="master"):
     """ Generates travis job body, that simply fails without
     running anything
     """
@@ -248,7 +248,7 @@ def generate_failure_callback():
     callback_body = {
         "request": {
             "message": "Systemtests failed. Build url:{}".format(triggered_by),
-            "branch": "master",
+            "branch": "{}".format(branch),
             "config": {
                 "merge_mode": "replace",
                 "sudo": "true",
@@ -279,7 +279,7 @@ if __name__ == "__main__":
 
     if args.failure:
         repo = adapters_info[args.adapter].repo
-        trigger_travis_build(generate_failure_callback(), args.owner, repo)
+        trigger_travis_build(generate_failure_callback(branch=args.branch), args.owner, repo)
     else:
         if args.test:
             job = generate_travis_job(args.adapter, args.owner, trigger_failure=False, branch=args.branch)
--- a/trigger_systemtests.py
+++ b/trigger_systemtests.py
@@ -103,7 +103,7 @@ def generate_travis_job(adapter, user, trigger_failure=True, branch="master"):
                                                                         tag=determine_image_tag())
 
     if trigger_failure:
-        after_failure_action += " python trigger_systemtests.py --failure --owner {USER} --adapter {ADAPTER}"
+         after_failure_action += " python trigger_systemtests.py --failure --owner {USER} --adapter {ADAPTER}" + "--branch {}".format(os.environ["TRAVIS_BRANCH"])
 
     # template for building this particular adapter
     build_template = {
--- a/trigger_systemtests.py
+++ b/trigger_systemtests.py
@@ -18,26 +18,27 @@ from collections import namedtuple
 
 adapter_info = namedtuple('adapter_info', 'repo tests base')
 
-adapters_info = {"openfoam": adapter_info('openfoam-adapter', ['of-of', 'of-ccx'],  'Ubuntu1604.home'),
-                "calculix":  adapter_info('calculix-adapter', ['of-ccx','su2-ccx'], 'Ubuntu1604.home'),
-                "su2":       adapter_info('su2-adapter',      ['su2-ccx'],          'Ubuntu1604.home'),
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
Exemplary usage of the feature implemented here is shown in https://github.com/precice/fenics-adapter/pull/60:

Excerpt from [.travis.yml](https://github.com/precice/fenics-adapter/blob/bd6a64d89c81dd7e8323bbe74f4fb5aa9ff58a6f/.travis.yml#L44-L45)
```
- curl -LO --retry 3 https://raw.githubusercontent.com/precice/systemtests/i_141/trigger_systemtests.py
- travis_wait 60 $PY trigger_systemtests.py --adapter fenics --wait --branch develop
```

trigger_systemtests.py
```diff
@@ -233,59 +238,56 @@ def trigger_travis_and_wait_and_respond(job_body, user, repo):
...
+    parser.add_argument('--wait', help='Whether exit only when the triggered build succeeds', action='store_true')
+    parser.add_argument('--test', help='Only print generated job, do not send the request', action='store_true')
+    parser.add_argument('--branch', help='Branch of systemtests being used for testing', default='master')
     args = parser.parse_args()
 
     if args.failure:
-        repo = adapters_info[ args.adapter ].repo
-        trigger_travis_build( generate_failure_callback(), args.owner,repo)
+        repo = adapters_info[args.adapter].repo
+        trigger_travis_build(generate_failure_callback(branch=args.branch), args.owner, repo)
```
Eder-K, line 282: This conditional branch that contains this line is only ever reached in the following case:

1. Some external repository, let's say fenics-adapter, commits something that triggers Travis to execute the `.travis.yml` file. This file contains the crucial line `python trigger_systemtests.py --adapter fenics`, which triggers a build on systemtests (and with this PR merged, on a specific branch of it). **Note that the `--wait` argument is omitted**, meaning that the fenics-adapter build on Travis will just send the request and instantly return as success.

2. A systemtests build is launched to run the test that was requested. If this test now fails we reach the after_failure_action
https://github.com/precice/systemtests/blob/55ffcc0bbe859639123f2faf93fbea292ad05a42/trigger_systemtests.py#L105-L106
So, the systemtests build will send a request **back to fenics-adapter** which originally triggered the whole sequence. 

At this point we have reached the highlighted line 282.

3. On fenics-adapter, a mock build is launched. This build will just fail and do nothing else, so that the actual feedback for a failed test is negative on the original repository as well.

https://github.com/precice/systemtests/blob/55ffcc0bbe859639123f2faf93fbea292ad05a42/trigger_systemtests.py#L282

Because of this, we have to be careful with which branch we choose here, as this must _not_ be a branch of systemtests. Currently, this mock build would always run on the master branch of the fenics-adapter. It would be nice to have this actually run on the specific branch that was originally commited to in point 1. Note that this is again a different branch than the systemtests branch that will be adjustable with this PR.

Eder-K: Since the mock request is generated on systemtests, we would need it to have received information in some way about which branch of the outside repository triggered it. One way might be to have the original request contain the following as after failure action
```
if trigger_failure:
    after_failure_action += " python trigger_systemtests.py --failure --owner {USER} --adapter {ADAPTER}" + "--branch {}".format(os.environ["TRAVIS_BRANCH"])
```
This way, when systemtests is executing `trigger_systemtests.py` as a result of a failed test, it will have received an `args.branch` value that correctly contains the branch of the outside repo that originally issued the test. Then we could actually call `generate_failure_callback(branch=args.branch)`

This way of formatting seems rather ugly, but it doesn't interfere with the formatting step done later on in

https://github.com/precice/systemtests/blob/55ffcc0bbe859639123f2faf93fbea292ad05a42/trigger_systemtests.py#L158-L160

We also don't require any further variables this way.

Suggestions?

trigger_systemtests.py
```diff
@@ -83,78 +83,81 @@ def determine_image_tag():
...
-                base_remote, tag = determine_image_tag())
+        {user}/{adapter}:{tag} --build-arg from={base_remote} .".format(adapter=
+                                                                                                adapters_info[
+                                                                                                    adapter].repo,
+                                                                        user=user, base_remote=
+                                                                                                base_remote,
+                                                                        tag=determine_image_tag())
 
     if trigger_failure:
         after_failure_action += " python trigger_systemtests.py --failure --owner {USER} --adapter {ADAPTER}"
```
Eder-K:
```suggestion
         after_failure_action += " python trigger_systemtests.py --failure --owner {USER} --adapter {ADAPTER}" + "--branch {}".format(os.environ["TRAVIS_BRANCH"])
```
See explanation above.

trigger_systemtests.py
BenjaminRueth: I'm obviously missing some context here. Let's try to resolve this issue in personal discussion.

PR #136, opened 2019-12-11
Use TRAVIS_BRANCH to select branch of toolchain being tested

BenjaminRodenberg:
closes #125 

# Todo

- [x] depends on https://github.com/precice/systemtests/pull/174. Has to be merged first.

PR #147, opened 2020-01-08
Overhaul pushing to precice_st_output

Eder-K:
Currently, after each test, we call `push.py` to commit some data related to the test to the repository [precice_st_output](https://github.com/precice/precice_st_output/tree/master). This comprises of an excerpt of the Travis job log, the job status and links to the related job/build.

Unfortunately, this is currently of very little use and most users probably just use the Travis webpage instead. This PR aims to improve the utility of precice_st_output in the following ways:

- [x] Change directory structure on precice_st_output to have each build quickly accessible (i.e. a seperate folder for each build with subfolders for each job)
- [x] Extract test output from containers and upload it to the related job folder on precice_st_output (these files are not accessible in other ways currently and are useful for use as reference output)
- [x] Store a full Travis log in a file
- [x] Store verbose console output of all participant containers for debug purposes (this is normally surpressed via `silent_compose.sh`)
- [x] Automatically generate a README in each build/job folder containing relevant information (possibly with jinja2)
- [x] Provide a README in the root directory of the repository explaining the file structure/folder numbering


Eder-K:
This PR is ready. Let's first fix the openfoam-adapter related tests by merging #174, then we can merge this one and the others that are finished (#136, #131)

PR #146, opened 2020-01-08
Fix FEniCS - FEniCS test

BenjaminRodenberg:
This PR still needs some changes in `tests/TestCompose_fe-fe.Ubuntu1804.home/Dockerfile.tutorial_data` as soon as https://github.com/precice/tutorials/pull/60 is merged:

* change `ARG branch=simplify-parallel-config` to `ARG branch=develop`
* change `RUN git clone --branch $branch https://github.com/uekerman/tutorials` to `RUN git clone --branch $branch https://github.com/precice/tutorials`

@Eder-K: if you want, you can also use the contents of this PR for your future big PR that fixes all the compatibility issues with the system tests that originate from https://github.com/precice/precice/pull/572.

Issue #139, opened 2019-12-16
Fix Ubuntu 16.04 home PETSc


BenjaminRodenberg:

BenjaminRodenberg:
Last successful run: https://travis-ci.org/precice/systemtests/builds/621545880

BenjaminRodenberg:
Building PETSc from source the [unit tests succeed](https://travis-ci.com/BenjaminRueth/systemtests/jobs/283877761). I created a [branch on my fork](https://github.com/BenjaminRueth/systemtests/tree/petsc_from_src) with a modified [`Dockerfile.Ubuntu1604.home`](https://github.com/BenjaminRueth/systemtests/blob/2178e8a789d132a9c1bf714540b50e942bc2ebf9/precice/Dockerfile.Ubuntu1604.home).

Note that building PETSc from source is not a solution but a workaround, since we want to be able to use PETSc provided via `apt` (and we should test it). However, it might be useful to provide an additional test, where PETSc is built from source. I would suggest to wait for the outcome of https://github.com/precice/precice/issues/667 before adding any new tests.

Eder-K:
With #222 , Ubuntu 16.04 is no longer tested. For newer Ubuntu versions, the PETSc builds now succeed with #246 .
Regardless of this, we do now successfully use PETSc from `apt` in 20.04 builds. Note that this is not possible in 18.04, as `apt` only provides PETSc v3.6, whereby we require at least PETSc v3.12. Due to this we continue to build PETSc from source in 18.04 builds.

Issue #115, opened 2019-10-24
Pushing logs to st_output fails due to invalid number of arguments


Eder-K:
Currently, when a systemtest triggered by an adapter attempts to push results to [precice_st_output](https://github.com/precice/precice_st_output), the push fails with an error message like in [this example](https://travis-ci.org/precice/systemtests/jobs/602015587#L564) (unwrap the push command if needed):

```bash
$ python push.py -s -t dealii-of
Cloning into '[secure]_st_output'...
EXECUTING: git clone https://github.com/[secure]/[secure]_st_output
EXECUTING: mkdir -p /home/travis/build/[secure]/systemtests/[secure]_st_output/Ubuntu1604.home
Traceback (most recent call last):
  File "push.py", line 161, in <module>
    add_job_log(args.test, not args.success, log_dir)
  File "push.py", line 116, in add_job_log
    create_job_log(systest, log, failed)
  File "push.py", line 70, in create_job_log
    *_, adapter_name, _, adapter_job_id = job_that_triggered.split('/')
ValueError: not enough values to unpack (expected at least 3, got 1)
```
This was not always the case. In tests like [this one](https://travis-ci.org/precice/systemtests/jobs/580208199#L511) from two months ago the push succeeds.

What is supposed to happen is that first the `trigger_systemtests.py` script sends a request that replaces the commit message on systemtests.

https://github.com/precice/systemtests/blob/14ba7f61133053c5d9afcf1af31441555fb8dbf0/trigger_systemtests.py#L160

For example: `fenics systemtest. Triggered by https://travis-ci.org/precice/fenics-adapater/jobs/602072485`.

The requested build will run the test and then execute `push.py`, whereby the script extracts information from the commit message that came with the request

https://github.com/precice/systemtests/blob/14ba7f61133053c5d9afcf1af31441555fb8dbf0/push.py#L65-L73

In current tests, however, the commit message read by `push.py` did not get overwritten and is simply the latest commit on the systemtests branch that gets used, see [this log](https://travis-ci.org/precice/systemtests/jobs/602072738#L568) (the final push command prints the full commit it parses before crashing, which matches the commit message of the [last commit](https://github.com/precice/systemtests/commit/c9dd71d24bb0152c85e26971e192c1f219fbd707) to this branch).

Therefore, the string that gets parsed doesn't include a link to the adapter job, which causes the operation to fail due to too few `/`-delimited parts.

Eder-K:
[...]

PR #148, opened 2020-01-09
Fix tests failing due to no output

Eder-K:
As mentioned in https://github.com/precice/systemtests/pull/146, a recent change to the precice.xml file (https://github.com/precice/precice/pull/572) caused most tests to crash due to outdated tutorials. After the PRs https://github.com/precice/tutorials/pull/60 and https://github.com/precice/openfoam-adapter/pull/107 have been merged, this PR should handle the remaining precice-config.xml file related issues.

This PR will contain fixes for each of the affected tests.

- [x] su2-ccx: 
    - [x] fails due to `Wrong attribute "distribution-type"`
    - [x] should work after merging https://github.com/precice/tutorials/pull/60.
    - [x] remove output redirect in ccx-container (https://github.com/precice/systemtests/pull/152#discussion_r365572195)
- [x] of-of: 
    - [x] `distribution-type` as well. Also depends on a line in the .xml file that probably must be adressed
    - [x] need merge of https://github.com/precice/openfoam-adapter/pull/107
- [x] of-of np: 
    - [x] `distribution-type` and additional line dependence, 
    - [x] need merge of https://github.com/precice/openfoam-adapter/pull/107
- [x] of-ccx: 
    - [x] `distribution-type` and additional line dependence.
- [x] fe-fe: 
    - [x] `distribution-type` see https://github.com/precice/systemtests/commit
    - [x] error in deployment phase currently.
    - [x] should work after merging https://github.com/precice/tutorials/pull/60./8d3fe2ed986c21919e787badcd7194fe3f288ec2
- [x] bindings/solverdummies:
    - [x] Currently still [fails](https://travis-ci.org/precice/systemtests/jobs/634663516#L676) due to `distribution-type`.
    - [x] error in deployment phase currently.
    - [x] Should work after merging https://github.com/precice/precice/pull/606. 
- [x] dealii-of:
    - [x] `distribution-type` and additional line dependence.
- [x] nutils-of:
    - [x] `distribution-type` and additional line dependence. (see #149 )
    - [x] error in deployment phase currently.
    - [ ] Hangs if the OF participant starts before the nutils one does.
- [x] [petsc] of-ccx: 
    - [x] fails as a consequence of [PETSc preCICE failing](https://travis-ci.org/precice/systemtests/jobs/638788539).


BenjaminRodenberg:
I would suggest to merge #146 into this PR. Then we address the problem in a single PR, which is a better approach from my point of view.

BenjaminRodenberg:
> I would suggest to merge #146 into this PR. Then we address the problem in a single PR, which is a better approach from my point of view.

Better idea: Since the (relevant) changes introduced in #146 are a one-liner, it is probably more appropriate to just manually apply these changes here, as well. I will take care of this and close #146.

BenjaminRodenberg:
I created a branch [`systemtests/hardcoded-fix-nooutput`](https://github.com/precice/systemtests/tree/hardcoded-fix-nooutput) that we can use for testing against the tutorials on [`uekerman/tutorials/simplify-parallel-config`](https://github.com/uekerman/tutorials/tree/simplify-parallel-config) until https://github.com/precice/tutorials/pull/60 is merged. 

**Important:** Don't merge [`systemtests/hardcoded-fix-nooutput`](https://github.com/precice/systemtests/tree/hardcoded-fix-nooutput) into this branch or into `develop`. It is only intended for testing purposes! We do not want to have the hardcoded branches in our final implementation

You can manually trigger a build of this branch on travis to test your implementation. I already triggered [a build](https://travis-ci.org/precice/systemtests/builds/635175668). I assume that the FEniCS-FEniCS test should work for that build, since it refers to the correct version of the tutorials.

BenjaminRodenberg:
The tests that I have updated, so far, still have problems with deployment. But I would suggest to first update all the tests such that they run through successfully and then worry about the deployment stage.

BenjaminRodenberg:
I checked [bindings/solverdummies](https://travis-ci.org/precice/systemtests/jobs/635278466) and they seem to work now (ignoring the usual deployment problem), since https://github.com/precice/precice/pull/606 has been merged.

BenjaminRodenberg:
Note that https://github.com/precice/openfoam-adapter/pull/107 is also relevant (for `of-of` and `of-of_np`)

Eder-K:
See https://github.com/precice/systemtests/pull/152#discussion_r365572195

For the SU2-Calculix test, I had to manually disable output in the ccx container so that the Travis log does not reach the size limit (which causes a job error):
https://github.com/precice/systemtests/blob/c2339986665561cd05dbe9845f54898caa0855cc/tests/TestCompose_su2-ccx/docker-compose.yml#L15

Before merging this PR, we should revert this change. it is only needed since we currently have each test output all of its container logs.

Eder-K:
> The tests that I have updated, so far, still have problems with deployment. But I would suggest to first update all the tests such that they run through successfully and then worry about the deployment stage.

The deployment error we see here is noted in https://github.com/precice/systemtests/issues/115 and relates to the commit message that is visible on the Travis job page (the text beside the branch name/PR number), it occurs whenever a build is manually triggered (on the webpage) or triggered by another repository. 
I think we don't need to worry about it, Travis actually does not support the mechanic that would be necessary for it to function (that is to have Travis read commit messages manipulated via API request or manual trigger). Besides, we will hopefully change to the new pushing system in https://github.com/precice/systemtests/pull/147 soon.

Eder-K:
The PETSc of-ccx test depends on the PETSc preCICE image, which is currently failing. This test will only work after https://github.com/precice/systemtests/issues/139 has been resolved.

BenjaminRodenberg:
> 
> 
> The PETSc of-ccx test depends on the PETSc preCICE image, which is currently failing. This test will only work after #139 has been resolved.

In this case I would suggest to allow failure on this test (within the scope of this PR) and fix it at a later point in time.

Then the last test we have to worry about it the nutil-of test. Correct?

MakisH:
@Eder-K I just now discussed with @BenjaminRueth and had a brief look over the nutils case.

A particularity of [this tutorial](https://github.com/precice/tutorials/tree/develop/CHT/flow-over-plate/buoyantPimpleFoam-nutils) is that it is supposed to start the solvers from inside the solvers' directories and `exchange-directory` is set to `..`.

Looking at the [compose file](https://github.com/precice/systemtests/blob/EderK-fix-nooutput/tests/TestCompose_nutils-of.Ubuntu1804/docker-compose.yml#L37-L39), we are wondering about the following:
- Why start the solvers from the home directory? Then we probably don't have access to `..` . @uekerman 
- Why setting a link? `ln -sf /home/precice/Data/Input/* .`
- Why did [this PR](https://github.com/precice/systemtests/pull/132) succeed? What changed in the meantime?

BenjaminRodenberg:
We want to merge this PR until Monday morning. See https://github.com/precice/precice/pull/614#issuecomment-577116110. If it is not 100% ready, we can fix the few failing tests later with a patch.

Eder-K:
The nutils-of test does occasionally work, but specifically when the openfoam participant is launching after the nutils one. I am still trying to find out why this happens, but as a **temporary** workaround setting a manual delay in the openfoam container does resolve the problem.

I am merging this 'hacky' version for now so that the develop branch finally runs the tests again. Meanwhile I will continue working on an actual for the nutils-of test here.

MakisH:
> I am merging this 'hacky' version for now so that the develop branch finally runs the tests again. Meanwhile I will continue working on an actual for the nutils-of test here.

Could you please document this in another issue then?

PR #152, opened 2020-01-12
Fix SU2-Calculix

Eder-K:
Update sed command to match the updated configuration files (https://github.com/precice/tutorials/pull/60).

### DIFF
--- a/tests/TestCompose_su2-ccx/docker-compose.yml
+++ b/tests/TestCompose_su2-ccx/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     command: >
             /bin/bash -c "ln -sf /home/precice/Data/Input Solid;
             ln -sf /home/precice/calculix-adapter/configs/* . &&
-            ccx_preCICE -i Solid/flap -precice-participant Calculix > systest_ccx.log 2>&1 &&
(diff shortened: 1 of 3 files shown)


tests/TestCompose_su2-ccx/docker-compose.yml
```diff
@@ -1,39 +1,39 @@
...
       - exchange:/home/precice/Data/Exchange/
       - output:/home/precice/Data/Output/
       - input_solid:/home/precice/Data/Input/
       - configs:/home/precice/calculix-adapter/configs
     command: >
-            /bin/bash -c "ln -sf /home/precice/Data/Input Solid; 
+            /bin/bash -c "ln -sf /home/precice/Data/Input Solid;
             ln -sf /home/precice/calculix-adapter/configs/* . &&
-            ccx_preCICE -i Solid/flap -precice-participant Calculix &&
+            ccx_preCICE -i Solid/flap -precice-participant Calculix > /dev/null 2>&1 &&
```
Eder-K, line 15: The test [succeeds](https://travis-ci.org/precice/systemtests/jobs/635921672?utm_medium=notification&utm_source=github_status), I had to temporarily disable output of the Calculix participant since the Travis log would get too large and the job errors. This is not critical, but before finally merging #148 (where we won't print the output by default) we should remove this again so it doesn't cause problems in the future.

PR #154, opened 2020-01-12
Fix dealii-of and ccx-of_fsi [PETSc]

Eder-K:


Eder-K:
Merging this, even though PETSc of-ccx [still fails](https://travis-ci.org/precice/systemtests/jobs/635936102?utm_medium=notification&utm_source=github_status). The underlying problem here is the [broken preCICE PETSc build](https://travis-ci.org/precice/systemtests/jobs/637887925?utm_medium=notification&utm_source=github_status), the changes made here should be sufficient though once the preCICE build is restored.

PR #129, opened 2019-11-28
Add tests for bindings on Ubuntu 18.04

BenjaminRodenberg:
Related to #98, since otherwise, we do not know that pip has to be installed from source on Ubuntu 18.04.

- [x] This PR should only be merged after #123 has been merged.
- [x] This PR depends on https://github.com/precice/precice/pull/606

PR #160, opened 2020-01-25
Add bindings test for Ubuntu 18.04

Eder-K:
Continuation of #129, after I merged it by accident.

The bindings 18.04 test currently fails due to a module error ([TravisCI link](https://travis-ci.org/precice/systemtests/jobs/641699804#L619)):
```
        from packaging import version
    ModuleNotFoundError: No module named 'packaging'
```


BenjaminRodenberg:
That's interesting, since we are currently trying to reproduce this kind of behaviour in https://github.com/precice/python-bindings/issues/21.

BenjaminRodenberg:
The `packaging` issue should be solved upstream: [`pyproject.toml`](https://github.com/precice/python-bindings/blob/develop/pyproject.toml#L3) defines the build time dependencies of `pyprecice`. I would still suggest to merge this PR, since it adds a systemtest for the bindings under Ubuntu 18.04 which is currently missing.

Eder-K:
Merged. The latest run for 18.04 bindings is indeed [successful](https://travis-ci.org/precice/systemtests/jobs/659772845?utm_medium=notification&utm_source=github_status) (ignoring the current issues with other tests) and the changes are not interfering with anything else.

PR #165, opened 2020-01-29
Identify requests from TravisCI website

Eder-K:
This small change should allow for triggering builds directly through the TravisCI website without always breaking the deployments during the test stage ([example](https://travis-ci.org/precice/systemtests/jobs/641747273#L2396)).

### DIFF
--- a/push.py
+++ b/push.py
@@ -66,13 +66,20 @@ def create_job_log(test, log, exit_status):
 
     # Decipher commit from the job message
     if (event_type == "api"):
-        job_that_triggered = commit_message.split("Triggered by:")[-1]
-        if job_that_triggered == "manual script call":
-           event_type = "manual trigger"
+        
+        # Check if api request came from triggering script
+        if ("Triggered by:" in commit_message):
+            job_that_triggered = commit_message.split("Triggered by:")[-1]
+            if job_that_triggered == "manual script call":
+                event_type = "manual trigger"
+            else:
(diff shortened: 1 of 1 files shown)


MakisH:
What exactly should I see in the example? What was the previous situation?

Eder-K:
In the deployment phase of a test, the push.py script is executed. This script fails if a build is manually launched from the TravisCI website: 
```
Cloning into '[secure]_st_output'...
EXECUTING: git clone https://github.com/[secure]/[secure]_st_output
EXECUTING: mkdir -p /home/travis/build/[secure]/systemtests/[secure]_st_output/Ubuntu1604.home
  File "push.py", line 170, in <module>
    add_job_log(args.test, not args.success, log_dir)
  File "push.py", line 119, in add_job_log
    create_job_log(systest, log, failed)
  File "push.py", line 73, in create_job_log
    *_, adapter_name, _, adapter_job_id = job_that_triggered.split('/')
ValueError: not enough values to unpack (expected at least 3, got 1)
Script failed with status 1
```

This happens because the script expects the commit message to be of the format
```
Triggered by: <link to repository that triggered build>
```
which is generally not the case when a build is manually triggered (we specify a custom commit message here). This format was intended for requests launched from the `trigger_systemtests.py` script only.

This PR makes this format requirement only apply if the commit message starts with `Triggered by: `. 
Note that this change will only be of short use, as the current push.py script should soon be outdated with the merging of #147 .

PR #172, opened 2020-02-03
Update reference results for su2-ccx test case

### LINKED ISSUE #171: Find reason for regression in SU-CCX test
There is a regression in the SU-CCX test case. See [here](https://travis-ci.org/precice/systemtests/jobs/645394815#L39780). Note that the difference between the reference results and the output is small, but we still should not completely ignore it. I am currently trying to find out what's the reason for this regression. Depending on the outcome the following two options exist for dealing with the regression:

* The regression is no problem -> Update the reference results
* The regression is a problem -> Try to fix the cause for the regression

BenjaminRodenberg:
closes #171 

Reference results need update due to change in Eigen version used under Ubuntu 16.04 from [version 3.3](https://packages.ubuntu.com/source/xenial/eigen3) (installed via `apt`) to [3.3.7](https://github.com/precice/systemtests/blob/18dbf9f66dcfb1eb27bf02b5fce63610b6ec35b8/precice/Dockerfile.Ubuntu1604.home#L26) (installed from source).


BenjaminRodenberg:
Tests work now. PR is ready for review.

Eder-K (approved):

PR #173, opened 2020-02-04
Check column names in compare_results.sh

### LINKED ISSUE #169: Tests do not validate first line (= column name) of output against reference?
The first line of the reference results (= column names) seems to have no influence on the tests. A recent breaking change updated some of the column names in the output. However, the tests do not validate the the first line of the output against the reference results. I came to this conclusion based on [this test](???), where the error (difference between output and reference) in the first line is not detected. As a reminder:

The first line of the reference results reads

https://github.com/precice/systemtests/blob/0b61ba36cce94a5b89e38963d3eebc970dbfd8a0/tests/TestCompose_nutils-of.Ubuntu1804/referenceOutput/Nutils/precice-Nutils-iterations.log#L1

However the format introduced in https://github.com/precice/precice/pull/620 differs from this one. In #168 I started updating the column names correspondingly for a different test case.

### What does this mean?

Currently our tests ignore the column names when comparing the output to the reference files. Therefore, we do not see errors in the column names. In the best case this leads to reference results that are inconsistent with our specification, if we forget to update the (e.g. via #168). This is ugly. In the worst case, columns might be totally mixed up and we do not notice it. This is bad.

Eder-K:
Closes #169.

Currently, the script creates a diff of the files in referenceOutput and Output, then filters the diff to extract only numerical fields. This is then used to compute the average and maximum differences.

This PR adds an additional filter parallel to the numerical filter to create two filtered diffs: one for numerical values (this is left unchanged) and one containing only lines with text. Both are evaluated separately.

This way, all text data is checked against the reference, **including version information**. This would also include timestamps, but none of the current tests has such values in their output data. The version comparison is somewhat useful, since we should be aware if there is a version difference between output and reference data. Furthermore, references are quickly updatable if this is the case.

_Additional comment:_ this PR also disables the printing of differing lines if a numerical difference is outside the tolerated range. This tends to not be helpful and more often than not clutters the Travis log up to the point of crashing the job due to too much output.


Eder-K:
[This](https://travis-ci.org/precice/systemtests/builds/650743820?utm_source=github_status&utm_medium=notification) is an example build using deliberately wrong column names for the fe-fe test.

compare_results.sh
```diff
@@ -55,16 +56,20 @@ if [ -n "$diff_files" ]; then
...
     # Filter output files, ignore lines with words (probably not the results)
     # removes |<>() characters
     # Do not delete "e", since it can be used as exponent
-    filtered_diff=$( echo "$rawdiff" | sed 's/(\|)\||\|>\|<//g; /[a-df-zA-Z]\|Version/d' )
+    num_diff=$( echo "$raw_diff" | sed 's/(\|)\||\|>\|<//g; /[a-df-zA-Z]\|Version/d' )
+    # Filter for text lines. Compare these seperately from numerical lines
+    text_diff=$( echo "$raw_diff" | sed '/[A-Za-df-z]/!d' )
+
 
     # Paiwise compares files fields, that are produces from diffs and computes average and maximum
```
BenjaminRodenberg:
```suggestion
    # Pairwise compares files fields, that are produces from diffs and computes average and maximum
```

BenjaminRodenberg (approved):
Looks good 👍

PR #177, opened 2020-02-15
Update reference data for of-ccx_fsi PETSc

### LINKED ISSUE #158: Fix [16.04 PETSc] OpenFOAM - Calculix test
This test is currently marked as allowed failure since it depends on the docker image that gets generated in the Ubuntu 16.04.home PETSc preCICE build (which currently fails #139).

This test should be looked at after #139 has been resolved.

Eder-K:
Closes #158 

This is a necessary reference file update after the upgrade from OpenFOAM v4 to v5 for all test containing OpenFOAM participants. Related to #174, which didn't include an update for this test.


Eder-K:
The test [now succeeds](https://travis-ci.org/precice/systemtests/builds/651047825?utm_source=github_status&utm_medium=notification), with the build error stemming from #179. This is ready to merge.

Issue #186, opened 2020-02-26
pip invocation warning in bindings test


fsimonis:
Test: bindings.Ubuntu1604.home

In step 13/35 : `RUN pip3 install --user .`

I receive the following warning:
```txt
Successfully built pyprecice mpi4py
Installing collected packages: mpi4py, numpy, pyprecice
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.
  WARNING: The scripts f2py, f2py3 and f2py3.5 are installed in '/home/precice/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed mpi4py-3.0.3 numpy-1.18.1 pyprecice-2.0.0.2
```

*edit:* For the first warning, refer to this issue. For the second warning, refer to #195.

BenjaminRodenberg:
>```
> WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
> Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
> ```

# Developer perspective

This warning can be avoided (at least locally on my system) by calling `pip` instead of `pip3`. We should carefully check whether `pip` refers to `pip2` or `pip3` in the setups that we use for testing before changing it.

Only because we can change it does not mean we should change it: It might be a good idea to intentionally keep this warning until it `will fail in a future version of pip`. If we keep using `pip3`, which will trigger the warning, our tests will tell us when this future version of pip has arrived and we have to update our documentation (see below).

# User perspective 

We should also make sure to update our python-related documentation correspondingly (we do not want to recommend using `pip3`, if this throws a warning). Here, it might be worth waiting a bit longer, since `pip3` always explicitly points to `pip` for `python3` while just calling `pip` might also refer to a `pip` for `python2` (see below). Note that we do not support `python2`. I also noticed that the warning is only thrown for quite recent `pip` versions (18.1, default on Ubuntu 19.10 does not give me the warning; 20.0.2 gives me the warning).

[...]

PR #194, opened 2020-02-27
Add comments pointing to #186

BenjaminRodenberg:
Like discussed in #186 using `pip3` will trigger a warning, if a recent `pip` version is used. However, https://github.com/precice/systemtests/issues/186#issuecomment-591887545 gives reasons to keep using `pip3` (at least for a while). This PR adds comments in the code to point developers to the corresponding issue.


BenjaminRodenberg:
@fsimonis do you think that the comments I added in the code are sufficient to help us bridging the time until `python3`/`pip3` is truly the standard and `python2`/`pip2` is truly dead? I would like to stick to using the more explicit calls to `python3`/`pip3` until that point in time.

PR #196, opened 2020-03-02
Download boost from sourceforge

BenjaminRodenberg:
The site that we previously used for downloading boost is not accessible anymore.

### DIFF
--- a/precice/Dockerfile.Ubuntu1604.home
+++ b/precice/Dockerfile.Ubuntu1604.home
@@ -28,7 +28,7 @@ ENV Eigen3_ROOT="/eigen/eigen-3.3.7"
 
 # Installing boost from source
 WORKDIR /boost-build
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg:
Note that the issue with https://dl.bintray.com/boostorg/release/1.65.1/source/boost_1_65_1.tar.bz2 not responding popped up quite recently. This was the first build where I could find the issue: https://travis-ci.org/precice/systemtests/builds/656624939?utm_medium=notification&utm_source=email

Eder-K:
Apparently the service is back up and the [latest cron job](https://travis-ci.org/precice/systemtests/builds/657245094) succeeded again.
Nontheless it should be worthwhile to switch sources, I assume that sourceforge would cause us less problems in the future.

Eder-K (approved):

PR #201, opened 2020-03-18
Extend log pushing to precice_st_output

Eder-K:
This PR extends the functionality of push.py such that we not only push logs from tests, but also from preCICE and adapter builds.


MakisH:
Just to be sure: We also need the solvers' log files (but not the results), e.g. `Fluid.log`, `Solid.log`.

Regarding the build logs, for the OpenFOAM adapter I would always like to have the `Allwmake.log`, `wmake.log`, `ldd.log`.

For the other adapters, I assume that the screen output is enough.

Eder-K:
This PR is ready to merge. Ideally, we should merge this after we find a fix for #211, though.

Eder-K:
#211 has been resolved by [including the necessary authorization token in a request header](https://github.com/precice/systemtests/pull/201/commits/cc21bd9531b9cd02986838134cee6b08d4823d12). Ready to merge.

PR #206, opened 2020-03-31
Add new path for dealii adapter in develop

davidscn:
According to our intention to add a new solver for deal ii, I rearranged the dealii-adapter structure a bit i.e I moved the linear solver to a separate directory and opened up a directory for the (upcoming) non-linear solver. 
AfaIk, the systemtests are triggered for the develop branches, hence we need to switch our travis path, to avoid failing tests. 
@uekerman Are you fine with [the new structure?](https://github.com/precice/dealii-adapter/tree/develop)
@Eder-K Is this all I need to do, to fix the systemtests?

### DIFF
--- a/adapters/Dockerfile.dealii-adapter
+++ b/adapters/Dockerfile.dealii-adapter
@@ -10,7 +10,7 @@ COPY --from=deal.ii /root/dealii/ /home/precice/dealii/
 USER precice
 WORKDIR /home/precice
 RUN git clone https://github.com/precice/dealii-adapter.git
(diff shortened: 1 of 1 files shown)


Eder-K (approved):
The code change looks good, that should be enough :+1: 

Though we should change the target branch to develop since master is quite outdated at this point.

uekerman:
> @uekerman Are you fine with [the new structure?](https://github.com/precice/dealii-adapter/tree/develop)

Yes, sounds reasonable :+1:

PR #209, opened 2020-04-03
Update the systemtests master branch

Eder-K:
The develop branch is in a relatively good state again, lets merge the newest version to master.


MakisH (approved):
We need this to actually check if adapter-specific system tests work. Let's merge this now and extend/fix it later if needed.

PR #205, opened 2020-03-27
Code_Aster-adapter system test

micheltakken:
Add the system test for Code_Aster - OpenFOAM.
The dockerfiles for Code_Aster Solver, and Code_Aster-adapter were added. 

Prepare the docker-compose.yml file (based on the OpenFOAM-OpenFOAM docker-compose.yml) for running a system test. Still needs some additions to run the Code_Aster-OpenFOAM tutorial.

PR #128, opened 2019-11-28
Add elastictube1d test

Eder-K:
Closes #16 .

Note that the implementation of the elastictube1d test introduced here is dependent on the tutorial structure used in https://github.com/precice/elastictube1d/pull/16 .

# Todo

- [x] `git revert bdc1b17` before merging (we want to test `develop`)

PR #212, opened 2020-04-10
Improve text filter in compare_results.sh

Eder-K:
Adjusts the sed filtering commands in `compare_results.sh` such that the text comparison is able to filter out timestamps. This is necessary for comparing results in the elastictube1d-cxx test #128 and the CodeAster test #205 (this might require additional adjustments).


tests/TestCompose_1dtube_cxx.Ubuntu1604.home/Dockerfile.1dtubecxx
```diff
@@ -17,4 +17,17 @@ RUN git clone https://github.com/precice/elastictube1d.git
 
 RUN mkdir -p Data/Output/Postproc
 WORKDIR /home/precice/elastictube1d/cxx
+<<<<<<< HEAD
+<<<<<<< HEAD:tests/Test_1dtube_cxx.Ubuntu1604.home/Dockerfile
```
MakisH: Merge conflict?

Eder-K: I missed some files by accident when merging, should be resolved now.

PR #220, opened 2020-05-05
Apply trigger_systemtest changes to master

Eder-K:
At this point, the triggering system is again be in a functioning state such that the sent requests correctly make use of the push script introduced in #147 ([example build on the deal.ii adapter](https://travis-ci.org/github/precice/systemtests/jobs/683603481)). This merge is the last necessary step so that the adapter repositories have access to the new version.


Eder-K:
Note that the build failure is due to TravisCI blocking docker image deployment from active pull requests, not from the code failing. The [deal.ii-adapter build mentioned above](https://travis-ci.org/github/precice/systemtests/builds/683603479) is representative of the actual code behavior.

MakisH (approved):

PR #228, opened 2020-06-05
Change namings to new dealii solver namings

davidscn:
For consistency, we needed to change the naming again according to the new dealii solver (https://github.com/precice/dealii-adapter/pull/25). We have now a test for the linear solver. 

The new dealii adapter has now two coupled solvers. We should add a test for the nonlinear solver as well in a future PR.

We changed `coupled_elasto_dynamics` -> `linear_elasticity`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ jobs:
     - name: "openFOAM adapter [PETSc] [Job failure permitted]"
     - name: "CalculiX adapter [PETSc] [Job failure permitted]"
     - name: "[16.04 PETSc] OpenFOAM <-> CalculiX [FSI] [Job failure permitted]"
-    - name: "[16.04] deal.ii adapter [Job failure permitted]"
     - name: "[16.04] SU2 <-> Calculix [unstable][Job failure permitted]"
 
   include:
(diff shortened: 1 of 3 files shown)


davidscn:
@Eder-K Which tutorials are currently tested? The current tutorials will fail additionally, because the individual output has changed.

Eder-K (approved):

davidscn:
The parameter file name changed as well, I added it. 
Still, the reference data for the tutorial needs to be refreshed.

Eder-K:
Currently, the only deal.ii related test we are running is FSI/flap_perp/OpenFOAM-deal.II.

davidscn:
Ok, can we test the 2D version instead? Should I provide reference data? Or do you want to open up a separate PR on this after the tutorials are merged?

davidscn:
@Eder-K We decided to switch to deal.II minimum version of 9.2.
Can we update the test?

Eder-K:
> Ok, can we test the 2D version instead? Should I provide reference data? Or do you want to open up a separate PR on this after the tutorials are merged?

It might be cleaner to add the other test in a separate PR, so I opened https://github.com/precice/systemtests/pull/232. If you already have existing reference data that would be great, though we might have to adjust it a little.

davidscn:
> If you already have existing reference data that would be great, though we might have to adjust it a little.

Ok, I will push to the other PR as soon as I have them. Do you want the `cout` or `vtk` results?

This PR can be merged as soon as the deal.II version is updated.

Eder-K:
VTK results would be best I think.

davidscn:
Ok, since the version has been updated and the tutorial test has a separate PR, this can be merged.

PR #230, opened 2020-06-11
Update master branch version of systemtests

Eder-K:
Several improvements have been made to the system for triggering tests from other adapters, we should merge them to master such that they come into effect (adapter repositories generally will use the master branch version of systemtests).

Apart from other minor bugfixes, this PR will also make the newest two tests (elastictube1d and OpenFOAM-CodeAster) run on the master branch.


MakisH (approved):
(not a proper review)

PR #232, opened 2020-06-18
Add deal.ii-of 2D test variant

Eder-K:


davidscn:
I see already some reference data, or am I missing something? 
Otherwise I will add deal.II vtk references for the flap_perp_2D.

Eder-K:
That is just a placeholder since I created this test from a copy of the 3D one.

Eder-K:
(missclicked the close PR button)

PR #246, opened 2020-07-16
Update PETSc to v3.12

Eder-K:
Building calculix-adapter with this version depends on https://github.com/precice/calculix-adapter/pull/41, otherwise ready to merge.

This PR updates the Ubuntu 18.04 and 16.04 preCICE builds to use PETSc v3.12.5.
As a result of this change, other changes have been performed as well:
- The builds for openfoam-adapter and calculix-adapter with PETSc enabled have been adjusted to run on Ubuntu 18.04 (previously 16.04). Other adapter builds (without PETSc) now have a 18.04 version.
- The docker image for the openfoam solver has been updated (previously we managed two different versions, they have been combined into one).
- The docker image for building the calculix-adapter now installs spooles and ARPACK via apt instead of building from source.
- The `branch` argument used for building docker images has been changed to specifically mean "branch of the preCICE repository". This is now separate from "branch of the adapter repository" (`adapter_branch`), as branch names for preCICE and adapters are generally not identical.
- Most tests are now operating on a 18.04 base (formerly 16.04). ~This is due to the tutorials requiring a more recent openmpi version (2.0 required, 16.04 only provides 1.10)~ (see comment by @MakisH below).


MakisH:
For documentation purposes, this point:

> - Most tests are now operating on a 18.04 base (formerly 16.04). This is due to the tutorials requiring a more recent openmpi version (2.0 required, 16.04 only provides 1.10).

seems to be inaccurate. Currently, all the cases are known to be working on 16.04 and OpenMPI 1.10.

However, in any case, let's just get rid of any 16.04 cases that cause trouble at the moment. We will soon change our whole infrastructure anyway and 16.04 will reach end-of-life in a few months.

PR #249, opened 2020-08-13
Update reference results for partitioned heat case

BenjaminRodenberg:
According to new feature introduced in https://github.com/precice/fenics-adapter/pull/83 and changes in https://github.com/precice/tutorials/pull/102.

It is expected that results change (https://github.com/precice/fenics-adapter/pull/82#issuecomment-672896370), but the change is only minimal and nothing to worry about.


IshaanDesai (approved):
Manually checked new reference results, all good :+1:

Eder-K (approved):

PR #97, opened 2019-09-18
add tests for Ubuntu 19.04

BenjaminRodenberg:
One of the biggest advantages of Ubuntu 19.04 is that we can use pip provided via `apt`. pip of Ubuntu 18.04 or older is not working with the python bindings. See [here](https://github.com/precice/systemtests/pull/96#issuecomment-532640406). Therefore, we should also consider testing Ubuntu 19.04.

# ToDo

- [ ] This PR should only be merged after #123 has been merged.
- [ ] Fix tests. See [here](https://github.com/precice/systemtests/pull/97#issuecomment-532661075)

PR #260, opened 2020-11-12
Install pip via apt

Eder-K:
Closes #97.

This changes the `pip` installation method: we now use `pip` provided by `apt`.

Note: the installation of `pip` should occur in the dockerfile that requires it, not in earlier base files. At the moment, only `fenics-adapter` requires `pip` to function, so it is the only adapter affected by this change.

In Ubuntu 18.04, we have found that the `pip` version provided by `apt` is too old, therefore we add the following step after installing `pip`:
```
pip3 install --user --upgrade pip
```

For context, the discussion below also mentions additional, unrelated changes to the dockerfiles that were initially attached to this PR. These changes have since moved to https://github.com/precice/systemtests/pull/269.


adapters/Dockerfile.fenics-adapter
```diff
@@ -18,32 +23,28 @@ RUN apt-get -qq update && apt-get -qq install \
     python3-dev && \
     rm -rf /var/lib/apt/lists/*
 
-# Install pip dependecies; we have to use pip, since cython provided by apt-get is too old.
-RUN wget -q https://bootstrap.pypa.io/get-pip.py -O get-pip.py && \
-    python3 get-pip.py
-
-
+# Install fenics
```
BenjaminRodenberg:
```suggestion
# Install FEniCS
```

adapters/Dockerfile.su2-adapter
```diff
@@ -1,10 +1,14 @@
+# Dockerfile for building the su2-adapter for preCICE
+# NOTE: Built for SU2 v6.0.0
+
 ARG branch=develop
-ARG from=precice/precice-ubuntu1604.home-${branch}
+ARG from=precice/precice-ubuntu1804.home-${branch}
```
BenjaminRodenberg: I'm not sure whether this kind of change should really go into this PR.

precice/Dockerfile.Ubuntu1604.home
```diff
@@ -1,5 +1,11 @@
 # Dockerfile for building preCICE on ubuntu 16.04
 
+# #################
+# WARNING: As of 11.2020, images using Ubuntu 16.04 are no longer maintained.
+#          Attempting to build this dockerfile or using it as base image for
+#          other builds is likely to fail.
+# #################
```
BenjaminRodenberg: See above. Not really sure whether this should go into this PR.

precice/Dockerfile.Ubuntu1804.home
```diff
@@ -14,63 +21,63 @@ RUN apt-get -qq update && apt-get -qq install \
...
+cd petsc-3.12.5 && \
+./configure && \
+make all test
 RUN chown -R precice:precice /petsc
 
-# Rebuild image if force_rebuild after that command
+USER precice
+
+# Upgrade pip to latest version
+RUN pip3 install --user --upgrade pip
```
BenjaminRodenberg: If we already provide the latest version of pip with this image, why do we install pip (again?) [here](https://github.com/precice/systemtests/pull/260/files?file-filters%5B%5D=.calculix-adapter&file-filters%5B%5D=.codeaster-adapter&file-filters%5B%5D=.dealii-adapter_3D&file-filters%5B%5D=.fenics-adapter&file-filters%5B%5D=.home&file-filters%5B%5D=.mpich&file-filters%5B%5D=.openfoam-adapter&file-filters%5B%5D=.package&file-filters%5B%5D=.su2-adapter&file-filters%5B%5D=.yml#diff-3a5025fdfc7d5ffbc4d1140e2ef5014804df9f0e376db4c02177e3933de08b79L43)?

BenjaminRodenberg: If I understand it correctly `python3-pip` is no preCICE dependency. Therefore, I would suggest to **not** install it in `precice/Dockerfile.Ubuntu1234.xyz`, but where we actually need pip the first time (bindings, fenics-adapter, code-aster). This allows us to make sure that preCICE does not (by accident) depend on pip.

BenjaminRodenberg (changes requested):
Using pip from apt is a good step. However, I have a few minor comments below.

precice/Dockerfile.Ubuntu1604.home
Eder-K: Fair point, I moved the comment cleanup related changes to a new PR (see https://github.com/precice/systemtests/pull/269), this PR now only handles the apt pip change. 

I force pushed my local branch here to keep the commit mess to a minimum, sorry about that :sweat_smile: At least the `git add -p` function proved to be very useful once more.

precice/Dockerfile.Ubuntu1804.home
Eder-K: I agree. In fact, it seems that `fenics-adapter` is currently the only adapter in need of pip. I removed the pip installs from the remaining precice dockerfiles and added the pip upgrade step to `fenics-adapter`.
It should be noted that if an adapter/test does require pip in the future, one should add `pip3 install --user --upgrade pip` as mentioned above to make sure that the version is sufficient at all times (currently needed for Ubuntu 18.04).

adapters/Dockerfile.fenics-adapter
Eder-K: Added.

adapters/Dockerfile.su2-adapter
Eder-K: See comments below.

BenjaminRodenberg (approved):
Looks good :+1:

PR #266, opened 2020-12-20
Update reference output of FEniCS-FEniCS test

IshaanDesai:
The reference output for the FEniCS-FEniCS test needs to be updated as the FEniCS-Adapter has a new design: https://github.com/precice/fenics-adapter/pull/96. To make the tutorials compatible with the new adapter design, several changes are done: https://github.com/precice/tutorials/pull/124. Both these changes slightly change the results of the test and hence the updating of reference output is necessary.


BenjaminRodenberg (approved):
There is even a slight improvement in the iterations. I opened an issue #266 on the need for updating the vtk output. However: Let's just update the vtks this time and worry about #266 later. Ready to merge :+1:

PR #268, opened 2020-12-21
Modifying reference output of FEniCS-FEniCS test

IshaanDesai:
In the previous PR updating the reference output of the FEniCS-FEniCS test: https://github.com/precice/systemtests/pull/266 the update was done by running the following configuration on a local machine:
* https://github.com/precice/fenics-adapter/commit/7dec20d7f492236b9b4f8843d8b9c17a2ea30328
* https://github.com/precice/tutorials/commit/1c891a7882ddf3a2abdfa50ecafb941f1395eebb

It was expected that once the locally updated results are updated as reference output, the system test will pass. That however did not happen. Further investigation reveals that the VTK output generated from the CI pipeline varies slightly with the VTK output obtained locally. The variation is close to machine precision (order of E-12). This variation can be a result of using different versions of dependencies in the CI pipeline. The variation is small enough to be ignored.

The following strategy is applied to update the VTK output of the FEniCS-FEniCS test:
* The results of the test are pushed to [precice_st_output](https://github.com/precice/precice_st_output) using the [push.py](https://github.com/precice/systemtests/blob/develop/push.py) script. By adding the `-o` or `--output` flag to the execution of this `push.py` script, the VTK output can be dumped in the `precice_st_output` repository. (@Eder-K thanks for this feature!)
* The VTK output dumped in `precice_st_output` is then copied as the reference output for the system test. This ensures that the output exactly matches.


BenjaminRodenberg (approved):
New results look good. Let's wait for the tests to finish and then merge :+1:

PR #269, opened 2020-12-22
Clean up precice/adapter Dockerfiles

Eder-K:
Adds/Updates comments in dockerfiles to more accurately describe their steps, purpose and current state. The comment format should now be consistent between files.

(Also fixes some weird indentations.)


MakisH:
Is there actually any formatting tool (e.g. clang-format) that also supports Dockerfiles? We could then integrate it to the repository.

PR #272, opened 2021-01-03
Update reference output of FEniCS-FEniCS test

IshaanDesai:
The bug: https://github.com/precice/fenics-adapter/issues/109 was found after `fenics-adapter` `v1.0.0` was released. The bug was not seen previously and hence wrong reference output was updated in the system test here: https://github.com/precice/systemtests/pull/268
This PR is a rectification after the bug was resolved.


tests/TestCompose_fe-fe/referenceOutput/Dirichlet/precice-output/DirichletNodes-HeatDirichlet.final.vtk
```diff
@@ -63,15 +63,15 @@ VECTORS VertexNormals float
...
+4.8200001397763668e+00
+5.0300001346197218e+00
+5.2999998400495185e+00
+5.6299999044847784e+00
+6.0200000931221211e+00
+6.4700000225566550e+00
+6.9800000432747424e+00
 7.5499999999999998e+00
 
 VECTORS Flux float
```
BenjaminRodenberg, line 77: Why is there no change in the Flux for the final result? This is a bit puzzling, but might have a good reason I do not see now.

BenjaminRodenberg (approved):
Looks good to me. Noteworthy change that unfortunately slipped our attention in #268: y-component of Flux is now correct, since the bug https://github.com/precice/fenics-adapter/issues/109 has been fixed upstream.

tests/TestCompose_fe-fe/referenceOutput/Dirichlet/precice-output/DirichletNodes-HeatDirichlet.final.vtk
IshaanDesai: It is even more puzzling that the flux in the `*.final.vtk` files in the [system test reference output](https://github.com/precice/systemtests/tree/develop/tests/TestCompose_fe-fe/referenceOutput/Dirichlet/precice-output) is zero in both X and Y directions. This is zero with the latest manual run of the tutorial too and hence there is no change.

As the values were unchanged from before only, perhaps it is safe to ignore this.

PR #273, opened 2021-01-20
Change "float" to "double" in reference output

BenjaminRodenberg:
This applies the changes introduced in https://github.com/precice/precice/commit/ffcb8e0d728234fc69063137867fd75083d94b46 to our reference results.


fsimonis:
I suppose the solution to the problem would be to compare the meshes using vtk or something like meshio.

BenjaminRodenberg:
> I suppose the solution to the problem would be to compare the meshes using vtk or something like meshio.

You mean we should entirely replace the [comparison script](https://github.com/precice/systemtests/blob/develop/compare_results.sh)? I totally agree. I could also contribute some python vtk skills, but I think this is a different issue and probably not a quick one.

BenjaminRodenberg:
Opened an issue #274.

MakisH (approved):
The change looks reasonable and I hope it is the only one needed.

Indeed we should use a VTK comparison tool for this instead, thank you for opening the issue. The reason we did not use this so far is because until recently we were testing the full output of all the solvers, which called for such a complicated comparison script (+ in our first attempt, the bash version was much faster than the python). The goal is to only test the VTK files exported by preCICE.

PR #275, opened 2021-01-20
Change "float" to "double" in reference output vol.2

BenjaminRodenberg:
Changes files missed in #273. Changing 738 files was obviously not enough.


BenjaminRodenberg:
Suggestion for review: Check the tree and see if the latest commit from this PR is missing somewhere. This was actually the approach I used for identifying the reference output that was still not updated.

MakisH (approved):
Difficult to review everything, but if something is still missing, the system tests will again complain.

PR #278, opened 2021-03-17
Remove 3D dealii OF test case

davidscn:
...which is not part of the tutorials any more. 

I could contribute some 'real' 3D cases in case we would like to test 3D cases of preCICE (or any of the solver).


davidscn:
For the sake of completeness: We can remove the '3D dealii docker' image as well. 1) we don't need it any more and 2) we can build the 3d version of the codes in the 2d image as well

MakisH (approved):

MakisH:
I guess you are referring to these images:
- https://hub.docker.com/r/precice/dealii-adapter_3d-ubuntu1804.home-release-v2.1.1
- https://hub.docker.com/r/precice/dealii-adapter_3d-ubuntu1604.home-develop

I can remove them as soon as this gets merged. But anyway, we probably need to do some clean-up in the images we have, when we start working on the system tests again.

davidscn:
I agree.

MakisH:
Update: I will do this when we merge also from `develop` to `master`. Otherwise, it will already start breaking.

But most probably I will do it the next time we massively clean-up.

PR #279, opened 2021-03-18
Update deal.II OpenFOAM reference data

davidscn:
I hope the tests run now on this PR.


davidscn:
@MakisH I will not update the 18.04 test case, since I have currently no image installed.

MakisH:
> @MakisH I will not update the 18.04 test case, since I have currently no image installed.

Then we should just keep this case as "allowed to fail" for now: https://github.com/precice/systemtests/blob/master/.travis.yml#L12-L19

I also don't have any 18.04 image ready, I would need to set it up.

davidscn:
I cannot see the current status (cancelled) of my updated test. How can I trigger the test (manually)?

MakisH:
You can either commit something, or start it from Travis: https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit (I hope you have permission, if not, let me know)

Currently, the adapters' Travis trigger the systemtests Travis, so it could be a bit more complicated than that.

davidscn:
_mhm_ I committed (obviously in this PR), but for me it looks as follows:
Screenshot from 2021-03-19 10-57-19

davidscn:
This is outdated.

